PR #2, opened 2016-03-15
Add topiary to the requirements

armish:
Trying to import this library fails due to missing dependencies:
screen shot 2016-03-15 at 5 27 16 pm


tavinathanson:
Whoops! Should `0.015` be `0.0.15`? Also, can you add a line break?

armish:
good catch! Just fixed it :+1:

PR #3, opened 2016-03-23
Add easy plotting for cohorts

tavinathanson:
`Plotting a KM curve is as simple as`cohort.plot_survival(snv_count)`. To be actually useful, I imagine it'll need a bunch more tweaking (e.g. easy filtering of the dataset).

Also added basic clinical data handling; it'll need more work, e.g. `pfs` isn't good enough to describe two different measures of progression-free survival.

### DIFF
--- a/cohorts/count.py
+++ b/cohorts/count.py
@@ -1,0 +1,66 @@
+# Copyright (c) 2016. Mount Sinai School of Medicine
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import numpy as np
+import pandas as pd
+
+from varcode import EffectCollection
+from varcode.effects import Substitution
+
+def snv_count(cohort):
+    sample_variants = cohort.load_variants()
+    def count_func(sample):
+        if sample in sample_variants:
+            return len(sample_variants[sample])
+        return np.nan
+    return count(cohort, count_func, count_col="snv_count")
+
+def nonsynonymous_snv_count(cohort):
+    sample_nonsynonymous_effects = cohort.load_effects(only_nonsynonymous=True)
+    def count_func(sample):
+        if sample in sample_nonsynonymous_effects:
+            return len(sample_nonsynonymous_effects[sample])
+        return np.nan
+    return count(cohort, count_func, count_col="nonsynonymous_snv_count")
+
+def missense_snv_count(cohort):
+    sample_nonsynonymous_effects = cohort.load_effects(only_nonsynonymous=True)
+    sample_missense_effects = dict(
+        [(sample,
+          EffectCollection(
+              [effect for effect in effects if type(effect) == Substitution]))
+         for (sample, effects) in sample_nonsynonymous_effects.items()])
+    def count_func(sample):
+        if sample in sample_missense_effects:
+            return len(sample_missense_effects[sample])
+        return np.nan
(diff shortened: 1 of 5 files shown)


cohorts/survival.py
```diff
@@ -0,0 +1,118 @@
+# Copyright (c) 2016. Mount Sinai School of Medicine
```
tavinathanson, line 1: Pulled straight from checkpoint trials; seems like a reasonable home for now, until we have a separate plotting util?

cohorts/plot.py
```diff
@@ -0,0 +1,85 @@
+# Copyright (c) 2016. Mount Sinai School of Medicine
```
tavinathanson, line 1: Also pulled straight from checkpoint utils

cohorts/survival.py
arahuja, line 1: Cool, I think it would be good to do that, with some standards on colors/size but this seems good for now

cohorts/count.py
```diff
@@ -0,0 +1,68 @@
...
+        if sample in sample_neoantigens["sample_id"].unique():
+            return len(sample_neoantigens[sample_neoantigens["sample_id"] == sample])
+        return np.nan
+    return count(cohort, count_func, count_col="neoantigen_count")
+
+def count(cohort, count_func, count_col):
+    df = cohort.clinical_dataframe.copy()
+    df[count_col] = df[cohort.clinical_dataframe_id_col].map(count_func)
+    original_len = len(df)
+    df = df[~df[count_col].isnull()]
```
arahuja, line 64: 1. I think you can use `notnull()` here 
2. Do we want to drop these columns or use the `np.nan` that is being set?

cohorts/count.py
```diff
@@ -0,0 +1,68 @@
...
+def neoantigen_count(cohort):
+    sample_neoantigens = cohort.load_neoantigens()
+    def count_func(sample):
+        if sample in sample_neoantigens["sample_id"].unique():
+            return len(sample_neoantigens[sample_neoantigens["sample_id"] == sample])
+        return np.nan
+    return count(cohort, count_func, count_col="neoantigen_count")
+
+def count(cohort, count_func, count_col):
+    df = cohort.clinical_dataframe.copy()
```
arahuja, line 61: Is this copy necessary?

cohorts/count.py
```diff
@@ -0,0 +1,68 @@
...
+            return len(sample_nonsynonymous_effects[sample])
+        return np.nan
+    return count(cohort, count_func, count_col="nonsynonymous_snv_count")
+
+def missense_snv_count(cohort):
+    sample_nonsynonymous_effects = cohort.load_effects(only_nonsynonymous=True)
+    sample_missense_effects = dict(
+        [(sample,
+          EffectCollection(
+              [effect for effect in effects if type(effect) == Substitution]))
```
arahuja, line 44: always curious is `type(x) == ...` preferred of `is instanceof`

arahuja:
@tavinathanson this looks good for the most part! I had some random questions here and there but didn't see anything wrong.

Some tests/comments would be nice, but I'm happy to contribute that too if you want to merge this in. I have some interface questions too (just how things are set up) but I think I'll need some experience using this as-is first.

cohorts/count.py
tavinathanson, line 64: We talked offline about 2: I don't want Mann Whitney plots, etc. to have to deal with NaNs. This does not permanently modify the underlying dataframe.

cohorts/count.py
tavinathanson, line 61: Yeah, I don't want to permanently modify the underlying dataframe.

cohorts/count.py
tavinathanson, line 44: Seems like `type` is the thing to use when you don't want inheritance to be factored in: http://stackoverflow.com/questions/1549801/differences-between-isinstance-and-type-in-python

tavinathanson:
As we talked about offline, I held off on tests and putting this in `hammerlab` until it seemed like it was actually being used. I figure that we can merge this in since plots already rely on it, and move it to `hammerlab` after adding tests, putting it on Travis, etc.

cohorts/count.py
tavinathanson, line 64: Fixed 1.

PR #4, opened 2016-03-25
Handle PFS better

tavinathanson:
Add a few assertions, and determine `progressed_or_dead` for PFS curves by ORing `progressed` and `dead`.

### DIFF
--- a/cohorts/load.py
+++ b/cohorts/load.py
@@ -47,6 +47,7 @@ class Cohort(object):
                  pfs_col=None,
                  dead_col=None,
                  progressed_col=None,
+                 progressed_or_dead_col=None,
                  hla_alleles=None,
                  cache_results=True,
                  snv_file_format_funcs=None,
@@ -62,6 +63,7 @@ class Cohort(object):
         self.pfs_col = pfs_col
         self.dead_col = dead_col
         self.progressed_col = progressed_col
+        self.progressed_or_dead_col = progressed_or_dead_col
         self.hla_alleles = hla_alleles
         self.cache_results = cache_results
         self.snv_file_format_funcs = snv_file_format_funcs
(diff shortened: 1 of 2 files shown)


cohorts/load.py
```diff
@@ -80,6 +86,18 @@ def __init__(self,
         self.nonsynonymous_effect_cache_name = "cached-nonsynonymous-effects"
         self.neoantigen_cache_name = "cached-neoantigens"
 
+    def verify_survival(self):
```
arahuja, line 88: This doesn't allow OS == PFS and it often will be when the values have been censored.

tavinathanson, line 88: How does it not allow for that? (In fact, I had a previous assert that at least some patients have OS == PFS, but took it out since it won't always be true.)

cohorts/load.py
```diff
@@ -299,7 +317,7 @@ def plot_survival(self, on, col=None, col_equals=None, how="os", threshold=None)
         results = plot_kmf(
             df=df,
             condition_col=plot_col,
-            censor_col=self.dead_col if how == "os" else self.progressed_col,
+            censor_col=self.dead_col if how == "os" else self.progressed_or_dead_col,
```
arahuja, line 329: Thinking about this a little more, this seems specific to a current use case. I think having the warning is great, but if there are two separate censor cols defined, it might be worth using the other sometimes?

tavinathanson, line 329: Good point. I'm curious when a separate PFS-specific censor column might result in different values than those inferred here. I believe that it could, depending on how that column got populated, but I'm fairly confused about that.

cohorts/load.py
arahuja, line 88: Sorry, I think I just read the assert wrong. Maybe the min is confusing me - it should be max?

cohorts/load.py
tavinathanson, line 329: (On second read, I'm not sure I'm correctly understanding this comment. Can you clarify what you meant by two separate censor cols?)

arahuja, line 329: I meant if someone did define an OS censor columns and PFS censor column, it seems like a good idea to alert when they are out of sync, but still use each censor column. Basically, I can't think of a case of why they'd be out of sync when it wasn't an error, in that case it seems worthwhile to explicitly fix the data? And if there ever was a use-case for that when it wasn't an error, this would override that

tavinathanson, line 329: Cool, that's what I thought you meant. Makes sense!

cohorts/load.py
tavinathanson, line 88: Yeah, I think you're right! I'm making it less confusing by using `all`.

cohorts/load.py
tavinathanson, line 329: Added an optional parameter for `progressed_or_dead` and check it against `dead | progressed` if it's passed in.

PR #5, opened 2016-03-30
Add stripboxplot

tavinathanson:
And use it in `plot_benefit`


cohorts/plot.py
```diff
@@ -18,6 +18,25 @@
 import seaborn as sb
 import pandas as pd
 
+def stripboxplot(x, y, data):
```
arahuja: maybe `kwargs` to striplot to adjust jitter/size/color etc?

arahuja:
Looks good! one suggestion

tavinathanson:
@arahuja :+1: added

PR #7, opened 2016-04-13
Fix KM plot axis

arahuja:
Similar to our discussion for other plots, the KM curve should pretty always range from 0 to 1 on the y-axis


tavinathanson:
LGTM! I'll actually use this within the next 1 minute.

PR #8, opened 2016-04-13
Adding comments to plotting functions

arahuja:
This add some docstrings to the two main plotting functions in `load.py`

Also, sets the `xlabel` for KM plots.

### DIFF
--- a/cohorts/load.py
+++ b/cohorts/load.py
@@ -17,9 +17,7 @@ from __future__ import print_function
 from os import path, makedirs
 from shutil import rmtree
 import pandas as pd
-import numpy as np
 import six.moves.cPickle as pickle
-import sys
 from types import FunctionType
 
 import varcode
@@ -29,7 +27,6 @@ from topiary import predict_epitopes_from_variants, epitopes_to_dataframe
 
 from .survival import plot_kmf
 from .plot import mann_whitney_plot, fishers_exact_plot
-from .count import count, snv_count, neoantigen_count
 
 class Cohort(object):
     """Represents a cohort of patients."""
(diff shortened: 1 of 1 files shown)


cohorts/load.py
```diff
@@ -307,7 +304,28 @@ def plot_init(self, on, col, col_equals):
...
+        odds-ratios and perform a Fisher's exact test.
+
+        If the variable is numeric, this will compare the distributions through
+        a Mann-Whitney test and plot the distributions with box-strip plot
+
+        Parameters
+        ----------
+        on : See `cohort.load.plot_init`
+        col : str, optional
+            If `on` is not specified, the column name on which to split the cohort
```
tavinathanson: This API is even confusing me already, so clearly could use improvement! 

That said, `on` is required, so this comment looks incorrect?

arahuja: Oh ok, I think I needed to read more of the `plot_init` docs, but I think I get it now. We are going to `plot_init` no matter what and `col` if it'll be a new column created from `on` based on `col_equals`?

tavinathanson: Yeah, basically; or `on` can just be a string name for a new column generated by comparing `col` with `col_equals`

tavinathanson:
Thanks! Looks good % minor comment

cohorts/load.py
arahuja: Hm, in the end it was a bit hard to succinctly state what `col` was. Gave it an attempt, but if you want we can just leave it as "Refer to `plot_init`"

tavinathanson:
LGTM :+1:

PR #11, opened 2016-04-25
Expressed neoantigens and other minor changes

tavinathanson:
In this PR:
- Add an `only_expressed` flag to `load_neoantigens` that uses isovar to grab protein sequences from RNA reads, and then sends those sequences through `mhctools`. CC @iskandr; am I doing anything obviously wrong?
- Add a `mannwhitneyu` test, and set two-sided to be the default.
- Return plot objects in plotting functions.

Note that I added a bunch of variables to `Cohort` that I'm not currently using anywhere, mostly for consistency. I need to find a better way to handle the `self.tumor_rna_bam_file_format_func(tumor_rna_bam_id)` situation; it's confusing that these format functions take in different sets of parameters.

### DIFF
--- a/cohorts/load.py
+++ b/cohorts/load.py
@@ -26,8 +26,11 @@ from types import FunctionType
 
 import varcode
 from varcode import VariantCollection, EffectCollection
-from mhctools import NetMHCcons
+from mhctools import NetMHCcons, EpitopeCollection
 from topiary import predict_epitopes_from_variants, epitopes_to_dataframe
+from topiary.sequence_helpers import contains_mutant_residues
+from isovar.protein_sequence import variants_to_protein_sequences_dataframe
+from pysam import AlignmentFile
 
 from .survival import plot_kmf
 from .plot import mann_whitney_plot, fishers_exact_plot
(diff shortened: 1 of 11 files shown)


arahuja:
> self.tumor_rna_bam_file_format_func(tumor_rna_bam_id) situation; it's confusing that these format functions take in different sets of parameters.

Yea, perhaps we can have more consistency there. This is really just an artifact of not having the `sample_id` with the RNA files originally so they were run with RNA ID that was available at the time. We could just rename the files to something more sane.

cohorts/load.py
```diff
@@ -263,19 +292,89 @@ def _load_single_sample_neoantigens(self, sample_idx, variant_type, merge_type,
...
+                    MAX_REFERENCE_TRANSCRIPT_MISMATCHES
+                )
+
+                # To ensure that 8-11mers overlap substitutions, we need at least this
+                # sequence length: (11 * 2) - 1
+                # Example:
+                # 123456789AB
+                #           123456789AB
+                # AAAAAAAAAAVAAAAAAAAAA
+                protein_sequence_length = (11 * 2) - 1
```
tavinathanson: I guess this should use the max peptide length vs. assuming 8-11mers

tavinathanson:
@arahuja I think this repo should work with RNA IDs that are != DNA IDs, though?

tavinathanson:
Going to open a separate PR with the latest changes.

iskandr:
My only suggestion would be to maybe increase the minimum MAPQ to 1 and if you're ultimately using the counts of neoepitopes with >= 3 spanning reads then this will be very sensitive to the degree of exonic coverage. Maybe normalize by number of reads mapping to exons?

PR #12, opened 2016-04-25
Update versioning to use versioneer

arahuja:
I've never been a fan of manually bumping the version on every commit. This change uses `versioneer` to manage this. 

`versioneer install` automatically creates this files and updates (except for updating `setup.py`). Versions are then per-commit which I think is great and releases are based on github tags. So if you make a change for example and check the verison it would show up as 

``` python
>>> cohorts.__version__
'0+untagged.60.g8f559e9'
```

I also updated the `.travis.yml` to only deploy tags (instead of releasing every push). So to release you could tag and push

```
> git tag <version>
> git push --tags
```

What do you think @tavinathanson? @iskandr can we use this on some of the other python packages as well?


arahuja:
> Coverage decreased (-6.3%) 

Sad, the autogen file really hurts us here

tavinathanson:
Nice! I lean towards thinking this is a good idea, including going with deploy-on-tag.

Kind of a bummer that we need to include the `versioneer` code in the repo; impacts our coverage, etc. Wonder if there's a way around that.

@iskandr Also curious to get your thoughts, mostly re PyPi deploy-on-tag vs. deploy-on-commit. I'm trying to remember why we went with the latter in general?

tavinathanson:
This sounds pretty hacky, but could we `urllib` download `versioneer` inside of `setup.py`?

arahuja:
@tavinathanson the autogen'd file has some variables filled based `setup.cfg` e.g. https://github.com/hammerlab/cohorts/pull/12/files#diff-e65acdc50d519281260d42812a02ffd1R45

Also, relying on internet to run `setup.py develop` wouldn't be great

arahuja:
We can exempt files from coveralls though, since really it shouldn't be used there http://stackoverflow.com/questions/18114126/is-there-a-way-to-exempt-files-folders-from-coveralls-test-coverage-scoring

tavinathanson:
Nice, exempting it from coveralls sounds like a much better approach.

tavinathanson:
Looks like @iskandr is moving towards tagged releases in isovar as well?

tavinathanson:
@arahuja Feel free to merge, ideally after resolving the coverage issue

arahuja:
@tavinathanson  added `.coveragerc` to exclude files from coverage analysis

Issue #1, opened 2016-03-15
Possible issue: sample_id as int vs str


tavinathanson:
Sometimes I see integer sorting and sometimes I see alphanumeric sorting; at the least, we should be sure that IDs always match up correctly (i.e. the right sample ID with the right BAM IDs).

arahuja:
I would support forcing the IDs to be strings on load in. Mostly because they may be 0-padded and if we interpret them as int they could lose their meaning. For example `00012`

PR #14, opened 2016-05-01
Refactor cohorts

tavinathanson:
Highlights:
- Instead of passing in a million arguments to a single `Cohort` object, split that out into `Patient`, `Sample`, etc. This also makes me less worried about the ordering of `sample_ids`, `tumor_bam_ids`, etc.; before, it would have been easier to accidentally re-order those lists IMO.
- Since a VCF path didn't make sense to be specified as a parameter of a `Sample` (where the `Sample` could be tumor or normal), I created `PairedSample`.
- A `Cohort` is a `Collection` of `Patients`.
- Move away from passing path functions around. Paths are now passed in at the time of `Sample` or `PairedSample` creation.
- I added logic for sending extra dataframes to the `Cohort` for joining; that's the `join_with` (which dataframe to join) and `join_how` (how to join). Dataframes get stored in a `Cohort` and can be joined with the rest of the data on demand. For example: `cohort = data.init_cohort(join_with=["pdl1", "tcr"], join_how="inner")` or `cohort = data.init_cohort(join_with="cibersort")`.
- I tried, wherever I could, to make the logic work for situations where patient ID != paired sample ID != sample ID, though I don't yet have tests for this. It's not really my priority right now; I just got roped into it as a result of the new objects.
- Move cufflinks loading to cohorts.
- The `isovar` stuff hasn't changed since https://github.com/hammerlab/cohorts/pull/11.
- Add a mannwhitneyu test, and set two-sided to be the default.
- Plotting functions return the plot for further manipulation.
- Fixes https://github.com/hammerlab/cohorts/issues/1 by forcing and asserting string IDs.

I tried to keep the consumption part of the API (vs. loading in the data) mostly the same; things this breaks:
- `self.clinical_dataframe` is now `self.as_dataframe()`
- Anything that touches the `Cohort` object.

Also note that some of the `CohortDataFrame` and `group_by` logic is a bit confusing. I'd like to simplify it and better document it at some point, but would prefer to defer to another PR on that.

### DIFF
--- a/cohorts/count.py
+++ b/cohorts/count.py
@@ -20,55 +20,49 @@ import pandas as pd
 from varcode import EffectCollection
 from varcode.effects import Substitution
 
-from .load import verify_group_by
-
 def snv_count(cohort, **kwargs):
-    sample_variants = cohort.load_variants(**kwargs)
-    def count_func(sample):
-        if sample in sample_variants:
-            return len(sample_variants[sample])
+    patient_variants = cohort.load_variants(**kwargs)
+    def count_func(patient_id):
+        if patient_id in patient_variants:
+            return len(patient_variants[patient_id])
         return np.nan
     return count(cohort, count_func, count_col="snv_count")
 
 def nonsynonymous_snv_count(cohort, **kwargs):
-    sample_nonsynonymous_effects = cohort.load_effects(only_nonsynonymous=True, **kwargs)
-    def count_func(sample):
-        if sample in sample_nonsynonymous_effects:
-            return len(sample_nonsynonymous_effects[sample])
+    patient_nonsynonymous_effects = cohort.load_effects(only_nonsynonymous=True, **kwargs)
+    def count_func(patient_id):
+        if patient_id in patient_nonsynonymous_effects:
+            return len(patient_nonsynonymous_effects[patient_id])
         return np.nan
     return count(cohort, count_func, count_col="nonsynonymous_snv_count")
 
 def missense_snv_count(cohort, **kwargs):
-    sample_nonsynonymous_effects = cohort.load_effects(only_nonsynonymous=True, **kwargs)
-    sample_missense_effects = dict(
-        [(sample,
+    patient_nonsynonymous_effects = cohort.load_effects(only_nonsynonymous=True, **kwargs)
+    patient_missense_effects = dict(
+        [(patient_id,
           EffectCollection(
               [effect for effect in effects if type(effect) == Substitution]))
-         for (sample, effects) in sample_nonsynonymous_effects.items()])
-    def count_func(sample):
-        if sample in sample_missense_effects:
-            return len(sample_missense_effects[sample])
+         for (patient_id, effects) in patient_nonsynonymous_effects.items()])
+    def count_func(patient_id):
+        if patient_id in patient_missense_effects:
+            return len(patient_missense_effects[patient_id])
         return np.nan
     return count(cohort, count_func, count_col="missense_snv_count")
 
 def neoantigen_count(cohort, **kwargs):
-    sample_neoantigens = cohort.load_neoantigens(**kwargs)
-    def count_func(sample):
-        if sample in sample_neoantigens["sample_id"].unique():
-            return len(sample_neoantigens[sample_neoantigens["sample_id"] == sample])
+    patient_neoantigens = cohort.load_neoantigens(**kwargs)
+    def count_func(patient_id):
+        if patient_id in patient_neoantigens["patient_id"].unique():
+            return len(patient_neoantigens[patient_neoantigens["patient_id"] == patient_id])
         return np.nan
     return count(cohort, count_func, count_col="neoantigen_count")
 
-def count(cohort, count_func, count_col, group_by="patient"):
-    verify_group_by(group_by)
-
-    id_col = "patient_id" if group_by == "patient" else "sample_id"
-
-    df = cohort.as_dataframe(group_by=group_by)
-    df[count_col] = df[id_col].map(count_func)
+def count(cohort, count_func, count_col):
+    df = cohort.as_dataframe()
+    df[count_col] = df["patient_id"].map(count_func)
     original_len = len(df)
     df = df[~df[count_col].isnull()]
     updated_len = len(df)
     if updated_len < original_len:
-        print("Missing count for %d samples: from %d to %d" % (original_len - updated_len, original_len, updated_len))
+        print("Missing count for %d patients: from %d to %d" % (original_len - updated_len, original_len, updated_len))
     return count_col, df
--- a/cohorts/count.py
+++ b/cohorts/count.py
@@ -57,9 +57,14 @@ def neoantigen_count(cohort, **kwargs):
         return np.nan
     return count(cohort, count_func, count_col="neoantigen_count")
 
-def count(cohort, count_func, count_col):
-    df = cohort.clinical_dataframe.copy()
-    df[count_col] = df[cohort.clinical_dataframe_id_col].map(count_func)
+def count(cohort, count_func, count_col, group_by="paired_sample"):
+    if group_by not in ["patient", "paired_sample"]:
+        raise ValueError("Invalid group_by: %s" % group_by)
+
+    id_col = "patient_id" if group_by == "patient" else "sample_id"
+
+    df = cohort.as_dataframe(group_by=group_by).copy()
+    df[count_col] = df[id_col].map(count_func)
     original_len = len(df)
     df = df[~df[count_col].isnull()]
     updated_len = len(df)
--- a/cohorts/count.py
+++ b/cohorts/count.py
@@ -20,6 +20,8 @@ import pandas as pd
 from varcode import EffectCollection
 from varcode.effects import Substitution
 
+from .load import verify_group_by
+
 def snv_count(cohort, **kwargs):
     sample_variants = cohort.load_variants(**kwargs)
     def count_func(sample):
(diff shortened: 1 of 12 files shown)


cohorts/load.py
```diff
@@ -372,24 +709,33 @@ def plot_survival(self, on, col=None, col_equals=None, how="os", threshold=None)
...
-    df = cohort.clinical_dataframe.copy()
+    df = cohort.as_dataframe()
     df[on] = df[col] == col_equals
     return on, df
+
+def verify_group_by(group_by):
+    if group_by not in ["patient", "paired_sample"]:
+        raise ValueError("Invalid group_by: %s" % group_by)
+
+def first_not_none(params, default):
```
arahuja: Didn't understand this usage above

cohorts/count.py
```diff
@@ -57,9 +59,13 @@ def count_func(sample):
...
-    df[count_col] = df[cohort.clinical_dataframe_id_col].map(count_func)
+def count(cohort, count_func, count_col, group_by="patient"):
+    verify_group_by(group_by)
+
+    id_col = "patient_id" if group_by == "patient" else "sample_id"
+
+    df = cohort.as_dataframe(group_by=group_by)
+    df[count_col] = df[id_col].map(count_func)
     original_len = len(df)
     df = df[~df[count_col].isnull()]
```
arahuja, line 64: Should we warn or alert when these are null

cohorts/count.py
```diff
@@ -57,9 +59,13 @@ def count_func(sample):
         return np.nan
     return count(cohort, count_func, count_col="neoantigen_count")
 
-def count(cohort, count_func, count_col):
-    df = cohort.clinical_dataframe.copy()
-    df[count_col] = df[cohort.clinical_dataframe_id_col].map(count_func)
+def count(cohort, count_func, count_col, group_by="patient"):
+    verify_group_by(group_by)
+
+    id_col = "patient_id" if group_by == "patient" else "sample_id"
```
arahuja: Some docs or comments on the group_by options here? From the PR description I'm gathering this separates samples like tumor and normal from a patient who has both?

cohorts/load.py
```diff
@@ -23,103 +23,320 @@
...
+    def __eq__(self, other):
+        if self is other:
+            return True
+        return (
+            self.__class__ == other.__class__ and
+            self.id == other.id)
+
+class CohortDataFrame(object):
+    """
+    Wraps a DataFrame with some information on how to join it.
```
arahuja: Not sure I understand this class

cohorts/load.py
```diff
@@ -23,103 +23,320 @@
...
+
+    def iter_paired_samples(self, include_parents=False):
+        for patient in self.elements:
+            for paired_sample in patient.paired_samples:
+                if include_parents:
+                    yield (patient, paired_sample)
+                else:
+                    yield paired_sample
+
+    def iter_tumor_samples(self, include_parents=False):
```
arahuja: include_parents?

cohorts/load.py
```diff
@@ -196,86 +410,206 @@ def _load_single_sample_variants(self, sample_idx, file_format_funcs, variant_ty
...
 
         if only_nonsynonymous:
             return nonsynonymous_effects
         return effects
 
-    def load_neoantigens(self, variant_type="snv", merge_type="union",
-                         epitope_lengths=[8, 9, 10, 11], ic50_cutoff=500,
-                         process_limit=10, max_file_records=None):
-        assert self.hla_alleles is not None, "Cannot predict neoantigens without HLA alleles"
+    def load_cufflinks(self, filter_ok=True):
```
arahuja, line 410: worth separating out cufflinks functionality?

tavinathanson, line 410: I don't understand your question: are you asking whether this should go in `cohorts` at all or whether it should be its own file in `cohorts`?

cohorts/load.py
tavinathanson: Yeah, sorry for the lack of comments. Wanted to get the PR out there but will update.

cohorts/load.py
arahuja, line 410: Was just thinking it's own file

tavinathanson, line 410: Sure, I think `load.py` is ripe for splitting up, though I generally like to make file splits separate PRs for ease of code reviewing.

tavinathanson:
@arahuja curious to get your thoughts on `Sample`, `PairedSample` and `Patient`: can these be more clearly named/organized? I think `Sample` vs. `PairedSample` is pretty confusing.

P.S. I may just get rid of a bunch of confusing logic and assume, for now, that every patient has just a single associated tumor and normal sample.

arahuja:
> curious to get your thoughts on Sample, PairedSample and Patient: can these be more clearly named/organized? I think Sample vs. PairedSample is pretty confusing.

Yea, it's hard to comment on the API without some experience using it. It did seem confusing at first, but I thought I'd wait to experience it first-hand a little. 

> I may just get rid of a bunch of confusing logic and assume, for now, that every patient has just a single associated tumor and normal sample.

I think that makes sense. I think that should always be the case. The time it might be confusing is we are likely to have an RNA BAM for tumor samples, but not normal (though we could). Similarly, if we are including TCRSeq data we have that on many samples (many normals and tumors possibly).  I haven't thought about it much, so whatever you think will be simplest for is probably right

tavinathanson:
@arahuja Why would it always be the case that patients would just have a single sample vs. samples at multiple timepoints/locations?

tavinathanson:
Simplified by:
- Getting rid of `PairedSample` and moving fields into `Patient`.
- Getting rid of all `group_by` logic.
- All IDs are `Patient` IDs.

test/data_generate.py
```diff
@@ -36,7 +36,7 @@ def generate_vcfs(id_to_mutation_count, file_format_func, template_name):
         template_path = data_path(template_name)
         vcf_reader = vcf.Reader(filename=template_path)
         file_path = generated_data_path(
-            path.join("vcfs", file_format_func(sample_id, None, None)))
+            path.join("vcfs", file_format % sample_id))
```
arahuja, line 39: worth documenting that `file_format` has a to be a format string? Why did you prefer this over the function?

cohorts/load.py
```diff
@@ -196,86 +373,204 @@ def _load_single_sample_variants(self, sample_idx, file_format_funcs, variant_ty
...
+                self.save_to_cache(df_isovar, self.cache_names["isovar"], patient.id, isovar_cached_file_name)
+
+            # Map from isovar rows to protein sequences
+            isovar_rows_to_protein_sequences = dict([
+                (frozenset(row.to_dict().items()), row["amino_acids"]) for (i, row) in df_isovar.iterrows()])
+
+            # MHC binding prediction
+            epitopes = mhc_model.predict(isovar_rows_to_protein_sequences)
+
+            # Only include peptides that overlap a variant; without this filter, when we use
```
arahuja: What is this comment referring to? What parameter controls this?

cohorts/load.py
```diff
@@ -23,131 +23,311 @@
...
-        cache_file = path.join(sample_cache_dir, file_name)
+        patient_cache_dir = path.join(cache_dir, str(patient_id))
+        cache_file = path.join(patient_cache_dir, file_name)
 
         if not path.exists(cache_file):
             return None
 
         if path.splitext(cache_file)[1] == ".csv":
-            return pd.read_csv(cache_file)
+            return pd.read_csv(cache_file, dtype={"patient_id": object})
```
arahuja, line 314: Maybe  special-case (and can be handled later) but `chromosome` always kills me this way. If there are no X, Y mutations or "chr" prefix pandas will load them as ints and kill any joins.

cohorts/load.py
```diff
@@ -23,131 +23,311 @@
...
+                cache_int_subdirs = set([int(name) for name in cache_subdirs])
+                if len(cache_subdirs) != len(cache_int_subdirs):
+                    bad_caches.append(cache_name)
+
+        if len(bad_caches) > 0:
+            raise ValueError("Caches %s have duplicate int/str directories" %
+                             str(bad_caches))
+
+    def as_dataframe(self, join_with=None, join_how=None):
+        join_dataframes = []
```
arahuja: Should this be used?

cohorts/load.py
```diff
@@ -14,7 +14,7 @@
 
 from __future__ import print_function
 
-from os import path, makedirs
+from os import path, makedirs, listdir
 from shutil import rmtree
 import pandas as pd
 import numpy as np
```
arahuja, line 20: unused

cohorts/count.py
```diff
@@ -21,48 +21,48 @@
...
         return np.nan
     return count(cohort, count_func, count_col="neoantigen_count")
 
 def count(cohort, count_func, count_col):
-    df = cohort.clinical_dataframe.copy()
-    df[count_col] = df[cohort.clinical_dataframe_id_col].map(count_func)
+    df = cohort.as_dataframe()
+    df[count_col] = df["patient_id"].map(count_func)
     original_len = len(df)
     df = df[~df[count_col].isnull()]
```
arahuja, line 64: This is not a change here, but I might prefer getting the dataframe with the NA values. Most numpy/pandas/seaborn functions are good about handling them and you don't lose the other info on the patient.  Can be handled elsewhere.

test/data_generate.py
tavinathanson, line 39: This is just for testing; shouldn't matter much? In general, I moved away from complex format functions to just setting the paths when creating the relevant objects.

cohorts/load.py
tavinathanson: Calling my `get_filtered_isovar_epitopes` does this, I'll make it more clear.

test/data_generate.py
arahuja, line 39: Ah ok sorry, didn't notice where it was.

cohorts/load.py
tavinathanson, line 314: Interesting, may be hard to handle here since that column could be under various names. I'll file an issue.

cohorts/load.py
tavinathanson: Good catch, old variable, removing.

cohorts/load.py
tavinathanson, line 20: It's used in `verify_cache`

cohorts/count.py
tavinathanson, line 64: Yeah, I agree this should be handled better; I'll file an issue?

tavinathanson:
@arahuja One annoyance is that sample_id to patient_id renders all existing cached variants/effects/neoantigens invalid. Is re-generating the cache more annoying than is tolerable?

arahuja:
No, I think that is fine, probably worth doing anyways with an update on all related tools?

tavinathanson:
@arahuja Cool, that was my original thinking.

PR #22, opened 2016-05-05
Fix dataframe loading bug

tavinathanson:
This does warrant a test. I'll write one up.

### DIFF
--- a/cohorts/load.py
+++ b/cohorts/load.py
@@ -203,7 +203,7 @@ class Cohort(Collection):
                  cache_results=True,
                  extra_df_loaders=[],
                  join_with=None,
-                 join_how="outer"):
+                 join_how="inner"):
         Collection.__init__(
             self,
             elements=patients)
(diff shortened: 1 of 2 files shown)


cohorts/load.py
```diff
@@ -267,8 +267,14 @@ def as_dataframe(self, join_with=None, join_how=None):
         if type(join_with) == str:
             join_with = [join_with]
 
+        # Convert strings to DataFrameLoader objects
+        df_loaders = []
+        for df_loader in self.df_loaders:
+            if df_loader.name in join_with:
+                df_loaders.append(df_loader)
+
```
arahuja, line 272: This could be 

```
df_loaders = [df_loader for df_loader in self.df_loaders if df_loader.name in join_with]
```

arahuja:
I'm not sure I know the bug here? And why the default changed to `inner` from `outer`

tavinathanson:
@arahuja the bug was that I was passing strings along to code that expected `DataFrameLoader` objects. Re `outer` to `inner`: I was being cautious with `outer` as the default, but it actually creates problems; namely, since outer joined patients don't have `OS` and `PFS`, they fail the `OS`/`PFS` verification. I do print the following when the join happens, though: `inner join with pdl1: 25 to 25 rows`

arahuja:
👍

PR #27, opened 2016-05-09
Clarify mannwhitney sidedness

tavinathanson:
Output sidedness to make sure we know how the `p-values` were generated.

@arahuja

### DIFF
--- a/cohorts/plot.py
+++ b/cohorts/plot.py
@@ -111,5 +111,13 @@ def mann_whitney_plot(data, condition, distribution,
         alternative=alternative
     )
 
-    print("Mann-Whitney test: U={}, p-value={}".format(U, pvalue))
+    if alternative is None:
+        raise ValueError("Must pick a mannwhitneyu alternative")
+    if alternative == "two-sided":
+        sided_str = alternative
+    else:
+        # alternative hypothesis: condition is 'less' or 'greater' than no-condition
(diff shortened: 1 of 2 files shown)


cohorts/plot.py
```diff
@@ -111,5 +111,13 @@ def mann_whitney_plot(data, condition, distribution,
...
 
-    print("Mann-Whitney test: U={}, p-value={}".format(U, pvalue))
+    if alternative is None:
+        raise ValueError("Must pick a mannwhitneyu alternative")
+    if alternative == "two-sided":
+        sided_str = alternative
+    else:
+        # alternative hypothesis: condition is 'less' or 'greater' than no-condition
+        op_str = ">" if alternative == "greater" else "<"
+        sided_str = "one-sided: %s %s not %s" % (condition, op_str, condition)
```
arahuja, line 121: What would this work out to? "one-sided: Benefit < not Benefit"? Is that what's intended? Sounds fine now that I think about it, but read weird to me originally (when I started typing this)

arahuja:
👍

cohorts/plot.py
tavinathanson, line 121: Yup, exactly. I never messed with the meaning of `less` vs. `greater` in my `scipy` PR, and it turns out my understanding of which means which was a little off. But based on the p values I see, this directionality appears correct.

Issue #15, opened 2016-05-01
Cache a dependency manifest


tavinathanson:
Every cached file can be accompanied by a `MANIFEST` that lists the software versions that created it.

tavinathanson:
Handled in a basic way by #28

PR #28, opened 2016-05-10
Add basic provenance tracking

tavinathanson:
Since I'm regenerating a shared cache right now, I figured I'd knock out some basic provenance tracking this time around.

What this does:
- Records the versions of a few key modules alongside each cached file.
- If those versions don't match up with the environment when loading from the cache, simply prints the diff for each patient.

Future work:
- Allow multiple cached versions to co-exist, keyed by the provenance. Currently, this is just an alert rather than a useful tool.
- Somehow track `NetMHC*` versions.
- Use `logging` rather than `print` statements and have a way to suppress and/or merge the many warnings.

### DIFF
--- a/cohorts/load.py
+++ b/cohorts/load.py
@@ -18,6 +18,7 @@ from os import path, makedirs, listdir
 from shutil import rmtree
 import pandas as pd
 import numpy as np
+import logging
 
 # pylint doesn't like this line
 # pylint: disable=no-name-in-module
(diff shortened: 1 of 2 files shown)


tavinathanson:
@armish I'm writing up a small test right now, and decided I should probably disable the warnings by default.

cohorts/load.py
```diff
@@ -313,12 +313,42 @@ def load_from_cache(self, cache_name, patient_id, file_name):
...
+            return set(dict(df.to_dict("split")["data"]).items())
+
+        def provenance_str_from_set(s):
+            return ["%s==%s" % (key, value) for (key, value) in s]
+
+        provenance_set = provenance_set_from_df(df_provenance)
+        provenance_set_previous = provenance_set_from_df(df_provenance_previous)
+        new_provenance_diff = provenance_set.difference(provenance_set_previous)
+        old_provenance_diff = provenance_set_previous.difference(provenance_set)
+        if new_provenance_diff != set():
```
armish: how about something like:

``` python
def check_diff(diff):
  if diff != set():
    print("Current provenance for patient %s: %s" % (
      patient_id, provenance_str_from_set(diff)))

check_diff(new_provenance_diff)
check_diff(old_provenance_diff)
```

to reduce the redundancy here?

cohorts/load.py
```diff
@@ -313,12 +313,42 @@ def load_from_cache(self, cache_name, patient_id, file_name):
...
+        def provenance_set_from_df(df):
+            return set(dict(df.to_dict("split")["data"]).items())
+
+        def provenance_str_from_set(s):
+            return ["%s==%s" % (key, value) for (key, value) in s]
+
+        provenance_set = provenance_set_from_df(df_provenance)
+        provenance_set_previous = provenance_set_from_df(df_provenance_previous)
+        new_provenance_diff = provenance_set.difference(provenance_set_previous)
+        old_provenance_diff = provenance_set_previous.difference(provenance_set)
```
armish: Why do you need this two-way diff'ing procedure? Is it because we might have each provenance having additional/missing keys in the future?

cohorts/load.py
```diff
@@ -313,12 +313,42 @@ def load_from_cache(self, cache_name, patient_id, file_name):
...
+                    old_provenance_diff)))
+
         if path.splitext(cache_file)[1] == ".csv":
             return pd.read_csv(cache_file, dtype={"patient_id": object})
         else:
             with open(cache_file, "rb") as f:
                 return pickle.load(f)
 
+    def generate_provenance(self, cache_name, patient_id):
+        df_provenance = pd.DataFrame()
```
armish: is this a data frame just because it is easier serialize/deserialize from the disk? 
Out of curiosity: are `Dict`/`JSON`s harder to save/load from the disk in Python?

armish:
nice one! :clap:

Just a few minor questions/comments for you. Feel free to merge it when/if you address/ignore them.

cohorts/load.py
tavinathanson: It was pure laziness and the ease of `df.to_csv`, probably should just output `JSON`!

cohorts/load.py
tavinathanson: Yeah, that was my thought: we might lose and gain modules.

cohorts/load.py
tavinathanson: I reduced the redundancy somewhat differently, let me know if it looks good?

tavinathanson:
Updates:
- Added a test.
- Added a `check_provenance` flag that defaults to `False`.

tavinathanson:
@armish Good to merge on green?

tavinathanson:
Fixes #15

armish:
Go for it! 

BTW: I think this looks much better with `df -> dict` switch. Thanks for doing that.

PR #29, opened 2016-05-10
Update isovar setup link

### LINKED ISSUE #23: Fix isovar git dependency
Not sure what the right way is to specify git packages in `setup.py`, but right now `pip install .` or `python setup.py install` give me the following message.

```
Collecting isovar>=0.0.2 (from cohorts==0.1.0+6.gd3244aa)
  Could not find a version that satisfies the requirement isovar>=0.0.2 (from cohorts==0.1.0+6.gd3244aa) (from versions: )
No matching distribution found for isovar>=0.0.2 (from cohorts==0.1.0+6.gd3244aa)
```

I can work around it with running the following first

``` sh
pip install git+git://github.com/hammerlab/isovar
```

arahuja:
Fixes #23  and also fixes the version to 0.0.3

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -64,10 +64,10 @@ if __name__ == "__main__":
             "topiary>=0.0.15",
             "six>=1.10.0",
             "lifelines>=0.9.1.0",
-            "isovar>=0.0.2",
+            "isovar==0.0.3",
         ],
         dependency_links=[
-            "git+git://github.com/hammerlab/isovar",
+            "https://github.com/hammerlab/isovar/archive/v0.0.3.tar.gz#egg=isovar-0.0.3",
         ],
         long_description=readme,
         packages=["cohorts"],


tavinathanson:
Thanks! Though probably best to pull the latest `isovar` vs. fixing it?

arahuja:
I thought fixing the version would just help us if isovar change API or functionality, we can make the decision to up it here?

Also, I can change the link to `https://github.com/hammerlab/isovar/tarball/master#egg=isovar-0.0.3` but I need to specify a version for it - what do you think?

tavinathanson:
I guess that makes sense re bumping `isovar` manually, to be better aware of what changes are happening under the hood.

So feel free to merge?

(Though I could imagine applying that logic to all libraries would be a pain. I've been mostly thinking about it from the other angle of assuming they will update and keeping track.)

arahuja:
> Though I could imagine applying that logic to all libraries would be a pain. I've been mostly thinking about it from the other angle of assuming they will update and keeping track.

Yea I think the difference is just that the other packages are well established so we shouldn't expect breaking changes in minor version increases (:cough: scipy) but isovar will likely move quickly and change APIs as needed in the early stage. Since it will directly impact the neoantigens it'd be good to understand them at each step and make sure all users stay in sync

PR #30, opened 2016-05-11
Fix RNA BAM path

tavinathanson:
A test would be a good idea, but not sure I need to block on that?

cc @arahuja

### DIFF
--- a/cohorts/load.py
+++ b/cohorts/load.py
@@ -537,6 +537,10 @@ class Cohort(Collection):
             if df_isovar is None:
                 import logging
                 logging.disable(logging.INFO)
+                if patient.tumor_sample is None:
+                    raise ValueError("Patient %s has no tumor sample" % patient.id)
+                if patient.tumor_sample.bam_path_rna is None:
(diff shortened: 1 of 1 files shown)


cohorts/load.py
```diff
@@ -537,7 +537,7 @@ def _load_single_patient_neoantigens(self, patient, variant_type,
             if df_isovar is None:
                 import logging
                 logging.disable(logging.INFO)
-                rna_bam_file = AlignmentFile(patient.rna_bam.path)
+                rna_bam_file = AlignmentFile(patient.tumor_sample.bam_path_rna)
```
arahuja, line 544: `tumor_sample` could be `None` worth checking? Not sure what to do in that case.

arahuja:
👍

cohorts/load.py
arahuja, line 544: As could `bam_path_rna` which seems more common

tavinathanson, line 544: Just added those checks 👍

PR #34, opened 2016-05-13
Pass kwargs through to function

tavinathanson:
Allow `on` functions with extra arguments.

The API can definitely be improved (`kwargs` are getting passed around haphazardly), but this is a quick fix for @e5c.


e5c:
LGTM

PR #36, opened 2016-05-18
Refactor how functions work

tavinathanson:
I was finding that `cohorts` `plot_*` functions weren't giving me all the functionality I wanted, while simultaneously giving me more than I needed.

In this PR:
- Got rid of confusing `col_equals` parameter everywhere, including supporting logic. It was intended to support the use case of `plot_benefit("is_met", "Met", "Y")`: i.e. the `"is_liver"` column would get created based on `"Liver Met" == "Y"`.
- Instead, I changed the format of functions in `cohorts` to simply take in a `row`. So now, if you want the above functionality: `plot_benefit(lambda row: row["Met"] == "Y", col="is_met")`.
- The `count` function is no longer a thing, so `count.py` --> `functions.py`.
- Renamed and upgraded `plot_init` to a more useful `as_dataframe`, while making the old `as_dataframe` the default behavior of the new one. The new `as_dataframe` can add/compute an arbitrary number of new columns. Example: `as_dataframe([snv_count, neoantigen_count])` gives you a `DataFrame` with both of those new columns, in addition to the default columns.
- Added `plot_joint` to take advantage of the multiple-function functionality. (It was also what drove me to add that functionality in the first place.)

More examples of what's possible now:

```
# Tell the cohort to automatically join on "pdl1" whenever `as_dataframe` is called. (This is not new.)
cohort = data.init_cohort(join_with="pdl1")

# Take advantage of the above to easily access a column from that joined-in data when plotting.
cohort.plot_benefit("IHC")
cohort.plot_benefit("IHC", col="better_ihc_name")

# Modify the value as you see fit.
cohort.plot_benefit(lambda row: row["IHC"] + 1)

# Use a function.
cohort.plot_benefit(neoantigen_count)
cohort.plot_benefit(neoantigen_count, only_expressed=True)
cohort.plot_benefit(lambda row: neoantigen_count(row, cohort, only_expressed=True))

# Get a dataframe that counts and merges. New columns "neoantigen_count" and "missense_snv_count" are created based on the function names.
cols, df = cohort.as_dataframe([neoantigen_count, missense_snv_count])

# Use your own function names.
cols, df = cohort.as_dataframe({"neoag": neoantigen_count, "snv": missense_snv_count})
cols, df = cohort.as_dataframe({"neoag": neoantigen_count, "ihc": lambda row: row["IHC"] + 1})

# Joint plot.
cohort.plot_joint({"neoag": neoantigen_count, "clonality": lambda row: row["Clonality"] + 1})
cohort.plot_joint(neoantigen_count, missense_snv_count)
```

cc @arahuja


tavinathanson:
@arahuja right now, `as_dataframe` returns either a `DataFrame` or a tuple of column name(s) and `DataFrame` (for when new columns are added). Is that weird? Open to better ideas.

setup.py
```diff
@@ -68,10 +68,9 @@
             "six>=1.10.0",
             "lifelines>=0.9.1.0",
             "scikit-learn>=0.17.1",
-            "isovar==0.0.3",
         ],
         dependency_links=[
-            "https://github.com/hammerlab/isovar/archive/v0.0.3.tar.gz#egg=isovar-0.0.3",
+            "git+git://github.com/hammerlab/isovar",
```
arahuja, line 73: Is this different now?

cohorts/load.py
```diff
@@ -395,7 +481,7 @@ def load_variants(self, variant_type="snv", merge_type="union"):
         assert variant_type in ["snv", "indel"], "Unknown variant type: %s" % variant_type
         patient_variants = {}
 
-        for patient in self:
+        for patient in self.iter_patients(patients):
```
arahuja, line 485: What does this provide over just `iter(patients)`?

cohorts/load.py
```diff
@@ -306,6 +307,78 @@ def as_dataframe(self, join_with=None, join_how=None):
...
+        if type(on) == str:
+            if col is not None:
+                df[col] = df[on]
+                return (col, df)
+            return (on, df)
+
+        def apply_func(on, col, df):
+            """
+            Sometimes we have functions that, by necessity, have more parameters
+            than just `row`. For a function like that, we assume that it has two
```
arahuja, line 346: This seems confusing to me ... but I don't have a better alternative

cohorts/load.py
```diff
@@ -306,6 +307,78 @@ def as_dataframe(self, join_with=None, join_how=None):
...
+            parameters: `row` and `cohort`. We use it to construct another function
+            that has only the `row` parameter so it can be sent to `DataFrame.apply`.
+            """
+            if col is None:
+                # Use the function name, or "column" for lambdas, if no
+                # name is provided for the newly created column.
+                col = on.__name__ if not is_lambda(on) else "column"
+            on_argcount = on.__code__.co_argcount
```
arahuja, line 354: Didn't know this existed

tavinathanson, line 354: Nor did I; not sure what the performance implications are, but not particularly worried for this use case.

cohorts/load.py
tavinathanson, line 346: Agreed :\

cohorts/load.py
tavinathanson, line 485: It handles the default of `patients=None` so that will use `self`; couldn't think of a cleaner way, since I can't put `self` as the default for `patients` in the method signatures.

setup.py
tavinathanson, line 73: Alex bumps the `isovar` version on every commit, and this fails whenever he does. That seems problematic, so I went back to the non-version-specific setup.

cohorts/count.py
```diff
@@ -1,68 +0,0 @@
...
-
-def neoantigen_count(cohort, **kwargs):
-    patient_neoantigens = cohort.load_neoantigens(**kwargs)
-    def count_func(patient_id):
-        if patient_id in patient_neoantigens["patient_id"].unique():
-            return len(patient_neoantigens[patient_neoantigens["patient_id"] == patient_id])
-        return np.nan
-    return count(cohort, count_func, count_col="neoantigen_count")
-
-def count(cohort, count_func, count_col):
```
tavinathanson, line 60: Note that this has been removed.

tavinathanson:
@arahuja Any more comments?

tavinathanson:
FYI @arman @eliza @jacki: I tried to make `cohorts` more usable. Check out the examples above.

(Also, note that some notebooks will break based on the changing definition of what functions like `neoantigen_count` look like.)

Aside from making certain things easier (e.g. a quick `jointplot` of mutation count vs. neoantigen count), the main reasons I think it's helpful to (continue to) stick our code in `cohorts`:
- Repeatability. @arahuja and I have run into several instances where we were counting or otherwise calling things in different ways (e.g. different de-duping strategies, missense vs. nonsynonymous, etc.), and it'd be great to know that all our analyses share the same basic strategies (and can all be easily updated when we change our minds). The [provenance](https://github.com/hammerlab/cohorts/pull/28) tracking will also hopefully prove useful for repeatability.
- Easily apply the same types of analyses to multiple projects.
- Keeping code public vs. siloed away.

Input/disagreements welcome!

arahuja:
@tavinathanson nothing else! It's a little weird to me that the count functions take a row and a cohort. It seems either they should functions of the cohort (we can load the necessary data and left-join) or a function of the row (compute some derived quantity). I'm sure it'll evolve and we have more use though

tavinathanson:
@arahuja Yeah, I hear you. That said: before this PR, I think the function API was even weirder. I like the idea of having one uniform way of creating a function, and I've already found it pretty valuable in my notebooks.

Definitely in favor of continuing to make it less confusing, though 👍

PR #39, opened 2016-05-20
Save variant metadata after merging variants

arahuja:


cohorts/load.py
```diff
@@ -489,38 +488,53 @@ def load_variants(self, patients=None, variant_type="snv", merge_type="union"):
...
-                combined_variants.append(set(variants.elements))
+            variant_collections = [
+                (vcf_path, varcode.load_vcf_fast(vcf_path)) 
+                for vcf_path in vcf_paths
+            ]
         except IOError:
             failed_io = True
 
-        if len(combined_variants) == 0 or failed_io:
+        if len(variant_collections) == 0 or failed_io:
```
tavinathanson: This will be undefined if there was an exception, no?

cohorts/load.py
```diff
@@ -489,38 +488,53 @@ def load_variants(self, patients=None, variant_type="snv", merge_type="union"):
...
             print("Variants did not exist for patient %s" % patient.id)
             return VariantCollection([])
 
-        if len(combined_variants) == 1:
+        if len(variant_collections) == 1:
             # There is nothing to merge
-            merged_variants =  VariantCollection(combined_variants[0])
+            merged_variants = VariantCollection(variant_collections[0][1].elements)
```
tavinathanson: Not obvious to me what the `[0][1]` is? Take the first `VariantCollection`, and then what?

tavinathanson: Oh, I see, `variant_collections` is a list of tuples. Not yet clear why the tuples are necessary?

arahuja: I wanted to hang to the vcf path's as well to know which file the metadata came from

cohorts/load.py
arahuja: Ah yea, I see that, will fix, can sub in `vcf_paths`

cohorts/load.py
arahuja: @tavinathanson I updated this to actually keep the metadata in the single vcf case as well

cohorts/load.py
arahuja: I switched the order of checks here

cohorts/load.py
```diff
@@ -489,38 +488,54 @@ def load_variants(self, patients=None, variant_type="snv", merge_type="union"):
...
-                merged_variants = VariantCollection(set.union(*combined_variants))
-            elif merge_type == "intersection":
-                merged_variants = VariantCollection(set.intersection(*combined_variants))
+            merged_variants = self._merge_variant_collections(dict(variant_collections), merge_type)
 
         self.save_to_cache(merged_variants, self.cache_names["variant"], patient.id, cached_file_name)
 
         return merged_variants
 
+    def _merge_variant_collections(self, variant_collections, merge_type):
```
tavinathanson: Confusing that `variant_collections` is a `dict` here and list of tuples above? Maybe rename to `vcf_to_variant_collections` here? Or make it a `dict` throughout?

tavinathanson:
LGTM % one more comment; though seems like `varcode` is a better place for logic to combine `VariantCollection`s? If you agree, maybe file an issue there that references this?

cohorts/load.py
arahuja: I renamed this `vcf_to_variant_collections` and rebased

arahuja:
I updated the variable name and rebased for the merge conflicts

> though seems like varcode is a better place for logic to combine VariantCollection

Maybe? I'm not sure it is, but I'm indifferent on it

tavinathanson:
👍

PR #45, opened 2016-05-27
Update building dataframe with additional data

### LINKED ISSUE #44: additional data assumes patient id field
When using `as_dataframe` the `additional_data` field from patients is merged back in, but assumes that one of keys is `patient_id`.

```
        additional_data_all_patients = defaultdict(list)
        for patient in self:
            if patient.additional_data is not None:
                for key, value in patient.additional_data.items():
                    additional_data_all_patients[key].append(value)

        if len(additional_data_all_patients) > 0:
            df = df.merge(pd.DataFrame(additional_data_all_patients), on="patient_id", how="left")

```

arahuja:
Fixes #44

### DIFF
--- a/test/test_basic.py
+++ b/test/test_basic.py
@@ -21,7 +21,7 @@ from cohorts import Cohort, Patient
 from cohorts.load import InvalidDataError
 
 import pandas as pd
-from nose.tools import raises, eq_
+from nose.tools import raises, eq_, ok_
 
 def make_simple_clinical_dataframe(
         os_list=None,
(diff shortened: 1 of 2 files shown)


test/test_basic.py
```diff
@@ -64,3 +67,10 @@ def test_progressed_vs_pfs():
 def test_simple_cohort():
     cohort = make_simple_cohort()
     eq_(len(cohort.as_dataframe()), 3)
+
+    columns = set(cohort.as_dataframe().columns)
+    assert "id" in columns
```
tavinathanson: I like to use `nose` functions in tests, e.g. `ok_()`, feel free to ignore.

arahuja: Updated to `ok_`

tavinathanson:
LGTM, code is cleaner too!

PR #48, opened 2016-05-27
Refactor plot_benefit

tavinathanson:
- `plot_boolean` is the generalized version of `plot_benefit`.
- `plot_benefit` now supports any name of the benefit col.

@arahuja


arahuja:
👍

PR #49, opened 2016-06-12
Add ability to filter variants

arahuja:


cohorts/load.py
```diff
@@ -518,7 +522,11 @@ def _load_single_patient_variants(self, patient, variant_type, merge_type):
 
         self.save_to_cache(merged_variants, self.cache_names["variant"], patient.id, cached_file_name)
 
-        return merged_variants
+        if filter_fn:
```
tavinathanson: Seems fragile to duplicate this code?

cohorts/load.py
```diff
@@ -485,24 +485,28 @@ def load_variants(self, patients=None, variant_type="snv", merge_type="union"):
         patient_variants = {}
 
         for patient in self.iter_patients(patients):
-            variants = self._load_single_patient_variants(patient, variant_type, merge_type)
+            variants = self._load_single_patient_variants(patient, variant_type, merge_type, filter_fn)
```
tavinathanson, line 489: Add `filter_fn` to the docstring?

tavinathanson:
@arahuja LGTM % comments

cohorts/load.py
arahuja, line 489: Will do

cohorts/load.py
arahuja: Sure, I can move it into a function above

arahuja:
@tavinathanson any idea what the Travis issue is?

tavinathanson:
@arahuja looks like two separate issues:
- https://travis-ci.org/hammerlab/cohorts/jobs/137107374: `pylint` issues; I guess `pylint` doesn't like what you're doing and you'll need to tell it to ignore those lines?
- https://travis-ci.org/hammerlab/cohorts/jobs/137107375: maybe this `pysam/cython` issue is due to some issue with the latest `isovar`? Wasn't it using `0.8.4` before?

arahuja:
@tavinathanson I addressed the comments above in the new the commit

Re the travis issues, I'll look into the first. For the second I think @alex recently bumped this version in `isovar`. We need to rely on master `isovar`  but not sure if it was released or how to specify that

tavinathanson:
I think we are currently relying on master `isovar` since we're pulling from GH and are not explicitely specifying a version?

arahuja:
@tavinathanson OK I followed `isovar`'s lead and added a line to have conda install an updated `pysam` 

```
conda install -c bioconda pysam=0.9.0
```

tavinathanson:
@arahuja cool; re `pylint`, here's an example disable: https://github.com/hammerlab/pyensembl/blob/22dd18ca3c8548c1d4e05c22da543097d2fdaae8/test/common.py#L45

arahuja:
@tavi hopefully Travis should work here, also added a similar function for filtering effect collections

tavinathanson:
👍

PR #50, opened 2016-06-13
Add mutect and strelka specific variant stats parsing

arahuja:


cohorts/variant_stats.py
```diff
@@ -0,0 +1,141 @@
...
+    SomaticVariantStats
+    """
+
+    all_stats = []
+    for (vcf, variant_metadata) in metadata.items():
+        if 'strelka' in vcf.lower():
+            stats = strelka_somatic_variant_stats(variant, variant_metadata)
+        elif 'mutect' in vcf.lower():
+            stats = mutect_somatic_variant_stats(variant, variant_metadata)
+        all_stats.append(stats)
```
tavinathanson, line 141: `else ValueError`?

arahuja, line 141: Good call, added and added a test that it raises properly.

cohorts/variant_stats.py
```diff
@@ -0,0 +1,141 @@
...
+    -------
+    SomaticVariantStats
+    """
+
+    sample_info = variant_metadata['sample_info']
+    # Ensure there are exactly two samples in the VCF, a tumor and normal
+    assert len(sample_info) == 2, "More than two samples found in the somatic VCF"
+
+    # Find the sample with the genotype field set to variant in the VCF
+    tumor_sample_infos = [info for info in sample_info.values() if info['GT'] == '0/1']
```
tavinathanson: Why not `sample_info["T1"]` like with Strelka?

arahuja: Unfortunately, that's not a standard. Mutect places the sample names as column headers. So this could really be anything, `Arun_T1/Arun_N1`, `blood\tumor` ... whatever the samples were named. In the test files, they just happen to be named `T1` and `N1`.

On the other hand, Strelka only uses two possible names, `NORMAL` or `TUMOR`

cohorts/variant_stats.py
```diff
@@ -0,0 +1,143 @@
...
+
+    # Ensure there is only one such sample
+    assert len(tumor_sample_infos) == 1, "More than one tumor sample found in the VCF file"
+
+    tumor_sample_info = tumor_sample_infos[0]
+    normal_sample_info = [info for info in sample_info.values() if info['GT'] != '0/1'][0]
```
tavinathanson: Why assert the `tumor_sample_infos` length but not the same check for `normal_sample_info`?

arahuja: There was an assert for exactly 2 samples above, and then one for exactly one tumor sample ... leaving exactly one left for normal? Should I add another one?

tavinathanson: I suppose not :)

tavinathanson:
Minor nit: been trying to use `"` everywhere for consistency. Feel free to ignore. Otherwise LGTM!

arahuja:
> Minor nit: been trying to use " everywhere for consistency. Feel free to ignore. Otherwise LGTM!

Will do, horrible habit of mine, always mixing

tavinathanson:
Talked offline: seems good to test the `merge_fn` functionality, @arahuja is going to add an overlapping variant to test that.

tavinathanson:
Thanks! LGTM round 2

PR #52, opened 2016-06-15
Return 0 for patients with no neoantigens

arahuja:
Removed the check of IDs since this give NaN for patients without neoantigens

### DIFF
--- a/cohorts/functions.py
+++ b/cohorts/functions.py
@@ -50,9 +50,7 @@ def missense_snv_count(row, cohort, **kwargs):
 def neoantigen_count(row, cohort, **kwargs):
     patient_id = row["patient_id"]
     patient_neoantigens = cohort.load_neoantigens(patients=[cohort.patient_from_id(patient_id)], **kwargs)
-    if patient_id in patient_neoantigens["patient_id"].unique():
-        return len(patient_neoantigens[patient_neoantigens["patient_id"] == patient_id])
-    return np.nan
+    return len(patient_neoantigens[patient_neoantigens["patient_id"] == patient_id])
 
 def expressed_neoantigen_count(row, cohort, **kwargs):
     return neoantigen_count(row, cohort, only_expressed=True)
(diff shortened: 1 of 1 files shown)


tavinathanson:
That was intended, and downstream the code checks for `null` and if `null` prints that patient X was skipped; which seems important when e.g. `cohort = data.init_cohort(only_patients_with_bams=False)` (sorry to reference a private repo here).

At a high level, it seems useful/necessary to maintain a distinction between `0` and not available? Where is the `NaN` a problem for you?

arahuja:
@tavinathanson in this case the correct answer is 0 though and not NaN (filtering a dataframe for an ID when there are 0 rows that match should be 0)

Compared to the other cases where we are checking if there is an entry corresponding to the ID and then  using `len`.  Here there are actually 0 neoantigens so there are 0 rows in the dataframe (and the ID is also not in the dataframe). Alternatively, we can add a separate check that we ran the neoantigens part for this ID.

Also, I haven't found this useful for the mutation counts either, since if a patient is missing VCF files, we return an empty `VariantCollection`.  Which means the check for the ID passes, and we (falsely) return 0 as opposed to NaN

tavinathanson:
Ah, I see. I guess we could look at what mutations are present to decide between 0 and `NaN`?

Re the empty `VariantCollection`: great point. It used to filter them out properly, so that's a bug.

tavinathanson:
Merge away per offline discussion:

> @tavinathanson: still not clear to me where you're running into issues with `nan`?
> @arahuja: When there are no neoantigens for example, `plot_benefit` etc throw out some of the patients
> @tavinathanson: suppose you can merge it in and i can follow it up with a PR that brings back `nan` in a non buggy way, sound reasonable?
> @arahuja: Sure - what were you thinking though?
> @tavinathanson: using load_variants to distinguish 0 from not available; and for not available, we should throw em out (not fully formed; does that make sense?)
> @arahuja: yea but it could be variants were available but hla wasn't
> @tavinathanson: can we rely on whether or not the neoantigen cache folder for that patient exists?
> @arahuja: not sure when it’s created (i.e. whether it’s created regardless and then just populated if anything is missing)
> @tavinathanson: i think not created regardless but have to check

PR #55, opened 2016-06-15
Allow extra keyword arguments in functions

tavinathanson:
Allows arbitrary arguments in functions. Example:

```
def snv_count(row, hello, **kwargs):
    return hello

cohort.as_dataframe(snv_count, hello="new message")
```

@arahuja

### DIFF
--- a/cohorts/functions.py
+++ b/cohorts/functions.py
@@ -39,8 +39,8 @@ def nonsynonymous_snv_count(row, cohort, **kwargs):
 def missense_snv_count(row, cohort, **kwargs):
     patient_id = row["patient_id"]
     patient_missense_effects = cohort.load_effects(
-        only_nonsynonymous=True, 
-        patients=[cohort.patient_from_id(patient_id)], 
+        only_nonsynonymous=True,
+        patients=[cohort.patient_from_id(patient_id)],
         filter_fn=lambda effect, variant_metadata: type(effect) == Substitution,
         **kwargs)
     if patient_id in patient_missense_effects:
--- a/cohorts/load.py
+++ b/cohorts/load.py
@@ -329,21 +329,19 @@ class Cohort(Collection):
         def apply_func(on, col, df):
             """
             Sometimes we have functions that, by necessity, have more parameters
-            than just `row`. For a function like that, we assume that it has two
-            parameters: `row` and `cohort`. We use it to construct another function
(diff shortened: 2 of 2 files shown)


arahuja:
@tavinathanson what happens if I have multiple functions?

tavinathanson:
@arahuja like this?

```
def snv_count(row, fn_1, fn_2, **kwargs):
    fn_1(2)
    fn_2(3)
    return 1

cohort.as_dataframe(snv_count, fn_1=lambda x: 2, fn_2=lambda x: 3)
```

That works, unless I'm misunderstanding the question?

arahuja:
@tavinathanson sorry I meant
`cohort.as_dataframe([snv_count, missense_snv_count])`

tavinathanson:
@arahuja looks like it applies them to both, e.g.:

```
def snv_count(row, fn_1, fn_2, **kwargs):
    fn_1(2)
    fn_2(3)
    return 1

def snv_count_2(row, fn_1, fn_2, **kwargs):
    fn_1(2)
    fn_2(3)
    return 1

cohort.as_dataframe([snv_count, snv_count_2], fn_1=lambda x: 2, fn_2=lambda x: 3)
```

And ends up in an error if they don't take the same parameters.

If you need to use two functions with different parameters, you can wrap the functions in lambdas vs. passing them via `as_dataframe` arguments.

Does that seem reasonable/expected?

tavinathanson:
Spoke with @arahuja offline; he suggested that passing kwargs to multiple functions is weird/unexpected.

@arahuja as of the last commit, this works:

```
cohort.as_dataframe({"snv_count": lambda row: snv_count(row, fn_1=lambda x: 2),
                     "snv_count_2": lambda row: snv_count_2(row, fn_1=lambda x: 3)})
```

This errors:

```
cohort.as_dataframe([snv_count, snv_count_2], fn_1=lambda x: 2, fn_2=lambda x: 3)
```

tavinathanson:
@arahuja good to merge?

arahuja:
@tavinathanson Did this always a allow a dict? Did not know that. The amount of options (list or func or dict) still seems weird, but if that is how it has been before, looks good.

tavinathanson:
It did, and the code is largely reused between `dict` and `list`. It's there to make it easier to specify the column names. Here, with just a list of `lambda`s, the columns would just default to `column_0` and `column_1`. Does that make sense?

PR #56, opened 2016-06-16
Fix bugs with caching and NaN handling

### LINKED ISSUE #26: Caching issues
Note the following scenario:
- Patient `2` removed from variant cache.
- Patient `2` underlying VCF file path deleted.
- `load_effects` called.
- Patient `2` doesn't exist; `print("Variants did not exist for patient %s" % patient.id)`
- `load_effects` called again, and nothing printed this time, because the `0` variants were cached in `load_effects`.

Long story short: errors need to be thrown so that we don't cache an error.

### LINKED ISSUE #54: Fix NaN situation
Described here: https://github.com/hammerlab/cohorts/pull/52#issuecomment-226265626

tavinathanson:
In this PR:
- Fix #54 by returning neoantigens as a dictionary of dataframes, more in line with how `load_variants` works, to track the difference between 0 and not-calculated neoantigens.
- Fix #26 by making sure we don't cache "0 effects" that are based on not-calculated/empty variants.
- Add `filter_not_null` to a few plotting functions where it was missing.
- Test 0 vs. NaN neoantigens by faking neoantigens in the cache.

cc @arahuja

### DIFF
--- a/cohorts/functions.py
+++ b/cohorts/functions.py
@@ -14,41 +14,57 @@
 
 from __future__ import print_function
 
+from .variant_filters import variant_qc_filter, effect_qc_filter, neoantigen_qc_filter
+
 import numpy as np
 from varcode.effects import Substitution
 
-def snv_count(row, cohort, **kwargs):
+def snv_count(row, cohort, qc_filter=False, **kwargs):
     patient_id = row["patient_id"]
     patient_variants = cohort.load_variants(
-        patients=[cohort.patient_from_id(patient_id)], **kwargs)
+        patients=[cohort.patient_from_id(patient_id)],
+        filter_fn=variant_qc_filter if qc_filter else None,
+        **kwargs)
     if patient_id in patient_variants:
         return len(patient_variants[patient_id])
     return np.nan
 
-def nonsynonymous_snv_count(row, cohort, **kwargs):
+def nonsynonymous_snv_count(row, cohort, qc_filter=False, **kwargs):
     patient_id = row["patient_id"]
     patient_nonsynonymous_effects = cohort.load_effects(
-        only_nonsynonymous=True, patients=[cohort.patient_from_id(patient_id)], **kwargs)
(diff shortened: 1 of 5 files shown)


cohorts/functions.py
```diff
@@ -14,40 +14,58 @@
 
 from __future__ import print_function
 
+from .variant_filters import variant_qc_filter, effect_qc_filter, neoantigen_qc_filter
+
 import numpy as np
 from varcode.effects import Substitution
 
-def snv_count(row, cohort, **kwargs):
+def snv_count(row, cohort, qc_filter=False, **kwargs):
```
arahuja, line 22: Any reason these default to `False` and not `None` directly?

cohorts/variant_filters.py
```diff
@@ -0,0 +1,53 @@
...
+
+import numpy as np
+from varcode import Substitution, Variant
+import re
+
+def variant_qc_filter(variant, variant_metadata):
+    somatic_stats = variant_stats_from_variant(variant, variant_metadata)
+
+    # Filter variant with depth < 30
+    if somatic_stats.tumor_stats.depth < 30 or somatic_stats.normal_stats.depth < 30:
```
arahuja, line 25: If moving this to `cohorts` should these be configurable?

cohorts/functions.py
tavinathanson, line 22: Seemed easier to say `qc_filter=True` than worry about which function to provide?

Separately, I think it probably makes sense to default to `True`, but that broke tests so I punted on it for now.

cohorts/variant_filters.py
tavinathanson, line 25: Figured I'd punt on that just to get it into the repo today?

tavinathanson:
@arahuja Good to merge or would you prefer I make it more customizable first?

arahuja:
No, looks good!

PR #61, opened 2016-06-21
Updates to filtering

tavinathanson:
In this PR:
- Add `normalized_per_mb` to all functions, defaulted to `True`.
- Change `qc_filter` in functions to `filter_fn` to be consistent with `load_*` functions.
- Make `filter_fn` in `load_*` (`load_variants`, `load_effects`, `load_neoantigens`) default to the appropriate filtering function.
- Updated required `topiary` version, as only recently does it include `chr`/`pos`/etc. in the output DataFrame, and we make use of that.
- Remove the regex string to variant hack, instead building a `Variant` object based on `chr`/`pos`/etc.
- Temporarily fix tests by having them not do filtering/normalization. (TODO later: both of these should be tested.)
- Add `filter_fn` to `load_polyphen*`.
- Update `min_mapping_quality` to `1`.

@arahuja

### DIFF
--- a/cohorts/load.py
+++ b/cohorts/load.py
@@ -41,8 +41,9 @@ from .survival import plot_kmf
 from .plot import mann_whitney_plot, fishers_exact_plot, roc_curve_plot
 from .collection import Collection
 from .varcode_utils import (filter_variants_with_metadata, filter_effects_with_metadata,
-                            filter_neoantigens_with_metadata)
-from .variant_filters import variant_qc_filter, effect_qc_filter, neoantigen_qc_filter
+                            filter_neoantigens_with_metadata, filter_polyphen_with_metadata)
+from .variant_filters import (variant_qc_filter, effect_qc_filter,
+                              neoantigen_qc_filter, polyphen_qc_filter)
 from . import variant_filters
 
 class InvalidDataError(ValueError):
@@ -555,7 +556,7 @@ class Cohort(Collection):
         return merged_variants
 
     def load_polyphen_annotations(self, as_dataframe=False,
-                                  variant_filter_fn=variant_qc_filter):
+                                  filter_fn=polyphen_qc_filter):
         """Load a dataframe containing polyphen2 annotations for all variants
 
         Parameters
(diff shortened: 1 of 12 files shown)


tavinathanson:
cc @jburos @e5c FYI

cohorts/load.py
```diff
@@ -568,26 +576,33 @@ def load_polyphen_annotations(self, as_dataframe=False):
...
         return patient_annotations
 
-    def _load_single_patient_polyphen(self, patient):
+    def _load_single_patient_polyphen(self, patient, filter_fn):
         cache_name = self.cache_names["polyphen"]
         cached_file_name = "polyphen-annotations.csv"
+        variants = self._load_single_patient_variants(patient,
+                                                      variant_type="snv",
+                                                      merge_type="union",
+                                                      filter_fn=None)
```
arahuja, line 595: Don't we want a `filter_fn` for the variants?

arahuja, line 595: Nevermind, I get what is happening here

tavinathanson, line 595: That gets handled in `filter_polyphen_with_metadata`; if we filtered variants here, we'd cache the filtered annotations vs. all of them. It's confusing.

cohorts/load.py
```diff
@@ -40,7 +40,10 @@
 from .survival import plot_kmf
 from .plot import mann_whitney_plot, fishers_exact_plot, roc_curve_plot
 from .collection import Collection
-from .varcode_utils import filter_variants_with_metadata, filter_effects_with_metadata
+from .varcode_utils import (filter_variants_with_metadata, filter_effects_with_metadata,
```
arahuja, line 43: I don't think the `filter_variants_with_metadata` or `effects is used

tavinathanson, line 43: They are, just not in these changes

arahuja, line 43: Nevermind it is, just not in this diff

arahuja:
Ignore my comments, looks good!

cohorts/load.py
arahuja, line 595: Yea I ran into the same thing on `load_effects`

Issue #66, opened 2016-06-28
VariantCollection metadata is lost after filtering


tavinathanson:
Example:

```
vc = cohort.load_variants()
vc[0].metadata
{}
```

```
vc = cohort.load_variants(filter_fn=None)
vc[0].metadata
{Variant(contig='1', ...
```

arahuja:
Good catch, looks like the metadata needs to be brought over from the older variant collection in `varcode_utils`

```
        return VariantCollection([variant
                                  for variant in variant_collection
                                  if filter_fn(variant, variant_collection.metadata[variant])
                                  ])
```

arahuja:
We should use varcode's `clone_with_new_elements`: https://github.com/hammerlab/varcode/blob/master/varcode/collection.py#L99-L107

tavinathanson:
Will `clone_with_new_elements` end up with extra metadata (for extra variants) that is no longer applicable? For my PR just now I manually copied over the metadata, though that seems non-ideal https://github.com/hammerlab/cohorts/pull/67/files#diff-481facac58e4fae6c10b3d93a889284eR576

arahuja:
You fixed this correct @tavinathanson?

tavinathanson:
Yup, fixed in https://github.com/hammerlab/varcode/issues/159

PR #67, opened 2016-06-28
Filter to expressed variants

tavinathanson:
In this PR:
- Factored some of the `df_isovar` logic out of `load_neoantigens` into a separate `_load_single_patient_isovar`, to use it outside the neoantigen context.
- Added `_filter_variants_to_expressed` to filter out any variant that isn't present in the `isovar` `DataFrame`.
- Randomly made it into this branch: parameterizing `variant_qc_filter`.

@arahuja

### DIFF
--- a/cohorts/load.py
+++ b/cohorts/load.py
@@ -385,7 +385,7 @@ class Cohort(Collection):
         # they apply to.
         if len(kwargs) > 0:
             raise ValueError("kwargs are not supported when collecting multiple functions "
-                             "as we don"t know which function they apply to.")
+                             "as we don't know which function they apply to.")
 
         if type(on) == dict:
             cols = []
@@ -598,8 +598,6 @@ class Cohort(Collection):
         cache_name = self.cache_names["polyphen"]
         cached_file_name = "polyphen-annotations.csv"
         variants = self._load_single_patient_variants(patient,
-                                                      variant_type=self.variant_type,
-                                                      merge_type=self.merge_type,
                                                       filter_fn=None)
         if variants is None:
             return None
--- a/cohorts/load.py
+++ b/cohorts/load.py
@@ -30,7 +30,7 @@ import vap  ## vcf-annotate-polyphen
 from sqlalchemy import create_engine
 
 import varcode
-from varcode import VariantCollection, EffectCollection
+from varcode import VariantCollection, EffectCollection, Variant
 from mhctools import NetMHCcons, EpitopeCollection
 from topiary import predict_epitopes_from_variants, epitopes_to_dataframe
 from topiary.sequence_helpers import contains_mutant_residues
@@ -470,7 +470,9 @@ class Cohort(Collection):
                 return patient
         raise ValueError("No patient with ID %s found" % id)
 
-    def load_variants(self, patients=None, variant_type="snv", merge_type="union", filter_fn=variant_qc_filter):
+    def load_variants(self, patients=None,
+                      variant_type="snv", merge_type="union",
+                      only_expressed=False, filter_fn=variant_qc_filter):
         """Load a dictionary of patient_id to varcode.VariantCollection
 
         Parameters
@@ -481,6 +483,8 @@ class Cohort(Collection):
             Load variants of a specific type, default 'snv'
         merge_type : {'union', 'intersection'}, optional
             Use this method to merge multiple variant sets for a single patient, default 'union'
+        only_expressed : boolean, optional
+            Only return expressed variants via isovar
         filter_fn : function
             Takes a variant and it's metadata and returns a boolean. Only variants returning True are preserved. Defaults to `variant_qc_filter`.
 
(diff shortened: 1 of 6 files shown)


cohorts/load.py
```diff
@@ -748,13 +800,52 @@ def load_neoantigens(self, patients=None, variant_type="snv", merge_type="union"
                 dfs[patient.id] = df_epitopes
         return dfs
 
+    def _load_single_patient_isovar(self, patient, variant_type, merge_type, epitope_lengths):
```
tavinathanson: I believe the content of this is all the same; just moved.

cohorts/load.py
```diff
@@ -533,8 +547,31 @@ def _load_single_patient_variants(self, patient, variant_type, merge_type, filte
...
+    def _filter_variants_to_expressed(self, variants, patient, variant_type, merge_type):
+        """Given a VariantCollection, filters that collection to only expressed variants."""
+        if len(variants) == 0:
+            return variants
+
+        expressed_variants = []
+        expressed_metadata = {}
+        # TODO: we're currently using the same isovar cache that we use for expressed
+        # neoantigen prediction; so we pass in the same epitope lengths.
+        # This is hacky and should be addressed.
```
tavinathanson: Open to ideas; wasn't sure how to best clean this up.

cohorts/load.py
```diff
@@ -533,8 +547,34 @@ def _load_single_patient_variants(self, patient, variant_type, merge_type, filte
...
+                                                     variant_type=variant_type,
+                                                     merge_type=merge_type)
+        for i, row in df_isovar.iterrows():
+            genome = variants[0].ensembl
+            variant = Variant(contig=row["chr"],
+                              start=row["pos"],
+                              ref=row["ref"],
+                              alt=row["alt"],
+                              ensembl=genome)
+            if variant in variants:
```
arahuja: This is going to go through all of `variants` each time, no?

cohorts/load.py
```diff
@@ -652,16 +697,26 @@ def load_effects(self, patients=None, only_nonsynonymous=False, variant_type="sn
...
+                                     variant_type, merge_type,
+                                     only_expressed, filter_fn):
+        def on_return(variants, effects):
+            if only_expressed:
+                variants = self._filter_variants_to_expressed(variants,
+                                                              patient=patient,
+                                                              variant_type=variant_type,
+                                                              merge_type=merge_type)
+                effects = [effect for effect in effects if effect.variant in variants]
+            return filter_effects_with_metadata(effects, variants.metadata, filter_fn)
```
arahuja: This works, but the first arg of `filter_effects_with_metadata` was supposed to be an `EffectCollection`

cohorts/load.py
tavinathanson: As opposed to caching the output? Yeah, I wanted to try to simplify by not caching this.

cohorts/variant_filters.py
```diff
@@ -20,26 +20,32 @@
 import pandas as pd
 from os import path
 
-def variant_qc_filter(variant, variant_metadata):
+DEFAULT_DEPTH = 30
```
arahuja: Cool!

cohorts/load.py
tavinathanson: Oh, maybe you mean that I'm iterating through the `VariantCollection` on each check; good point.

arahuja:
My thoughts on this are I'm not sure it's worth building in the isovar filtering as special case to the `load_*` functions and leaving it as something to do after the fact or making a `filter_fn` that could handle it.

tavinathanson:
@arahuja curious why not?

cohorts/load.py
tavinathanson: Good catch, I blame Python.

tavinathanson:
We spoke offline; @arahuja clarified that the `cohorts.varcode_utils` functions are more general than I realized, so I can indeed plop this logic in there.

tavinathanson:
@arahuja I'm not entirely clear on the best way to do this, since my `variant_expressed_filter` and `effect_expressed_filter` need more than `variant`/`effect` and `variant_metadata` (the `cohort`, `patient`, `variant_type`, `merge_type`). I could put `variant_expressed_filter` inside a count function and make use of the closure, but that's not easily passed to `load_variants`.

Perhaps I should replace `variant_metadata` with heavier-weight object that has the `VariantCollection` metadata as well as `cohort`, `patient`, etc.?

arahuja:
> Perhaps I should replace variant_metadata with heavier-weight object that has the VariantCollection metadata as well as cohort, patient, etc.?

Perhaps? I think needing cohort would be odd, but it somewhat shows that some of the `load_single` functions should probably be pushed into `Patient`

tavinathanson:
@arahuja I agree re pushing more functions to `Patient`; for another time, though :)

tavinathanson:
12 days later, some updates to this PR:
- Create `FilterableVariant`, `FilterableEffect`, etc. as a way to cleanly specify `filter_fn`s as single-parameter functions. Also removes the need for `neoantigen_qc_filter` vs. `variant_qc_filter` since `FilterableNeoantigen` is a `FilterableVariant`.
- Use the above to handle expression filtering instead of adding `only_expressed` everywhere.
- Fix https://github.com/hammerlab/cohorts/issues/73.
- Fix https://github.com/hammerlab/cohorts/issues/66.
- Move `merge_type` and `variant_type` into `Cohort` so they aren't passed around to a million methods.
- Add additional parameters to `variant_qc_filter`.

Back to you @arahuja

cohorts/load.py
```diff
@@ -215,10 +223,16 @@ def __init__(self,
                  join_how="inner",
                  check_provenance=False,
                  polyphen_dump_path=None,
-                 pageant_coverage_path=None):
+                 pageant_coverage_path=None,
+                 variant_type="snv",
```
arahuja, line 231: any reason this is a cohort level property now?

cohorts/load.py
```diff
@@ -360,7 +378,7 @@ def is_lambda(func):
         if type(on) == FunctionType:
             return apply_func(on, col, df)
 
-        # For multiple functions, don't allow kwargs since we won't know which functions
+        # For multiple functions, don't allow kwargs since we won"t know which functions
```
arahuja: weird quote change

cohorts/load.py
tavinathanson, line 231: Yeah, it became cumbersome and wordy to pass it around everywhere. Agree it's a little weird--

arahuja, line 231: Yea I think I prefer the alternative, but also see the issue that you can pass one to `load_variants` and  a different set to `load_effects` which could be confusing, so I'm fine with this

cohorts/load.py
tavinathanson: Fixed

PR #68, opened 2016-07-06
Summarize provenance

jburos:
Added two functions to Cohort class, with a goal of allowing a high-level summary of provenance in data reports (such as ipynbs). Welcome feedback on the implementation of this.

Functions include: 
1. `summarize_provenance`: 
   - tests for equivalence in provenance among all patients in each cache dir
   - summarizes provenance (if equivalent among all patients) for each cache dir
   - (prints warning if per-patient cache not equivalent within a cache_dir)
2. `summarize_unique_provenance`:
   - summarizes per-cache-dir provenance, suitable for display
     - if equivalent among all cache dirs, prints single provenance
     - otherwise, prints per-cache-dir provenance  

Both functions include among provenance files a hash of the `as_dataframe` object, which can be used to identify if two analyses are using the most recent data.

Example output: 

```
> cohort.summarize_unique_provenance()
Out[2]:
{u'cohorts': u'0+untagged.188.g6260531.dirty',
 u'dfhash': -8604513298553306351,
 u'isovar': u'0.0.6',
 u'mhctools': u'0.2.3',
 u'numpy': u'1.11.0',
 u'pandas': u'0.18.1',
 u'pyensembl': u'0.9.1',
 u'scipy': u'0.17.0',
 u'topiary': u'0.0.21',
 u'varcode': u'0.4.14'}
```


cohorts/load.py
```diff
@@ -1037,3 +1124,33 @@ def filter_not_null(df, col):
     if updated_len < original_len:
         print("Missing %s for %d patients: from %d to %d" % (col, original_len - updated_len, original_len, updated_len))
     return df
+
+def _provenance_str(provenance):
```
tavinathanson, line 1166: This is now a duplicate of `provenance_str` in `load_from_cache`, right? Can we get rid of that one?

cohorts/load.py
```diff
@@ -1037,3 +1124,33 @@ def filter_not_null(df, col):
...
+    """ utility function used by _compare_provenance to print diff
+    """
+    return ["%s==%s" % (key, value) for (key, value) in provenance]
+
+
+def _compare_provenance(provenance, comparison):
```
tavinathanson: Can we use this in the `check_provenance` section of `load_from_cache`?

tavinathanson: Maybe call these variables `this_provenance` and `other_provenance`?

cohorts/load.py
```diff
@@ -1020,6 +1020,93 @@ def plot_joint(self, on, on_two=None, **kwargs):
...
+            cache_dirs.
+
+        """
+        provenance_summary = {}
+        df = self.as_dataframe()
+        df_hash = hash(str(df.sort_values('patient_id')))
+        for cache in self.cache_names:
+            cache_name = self.cache_names[cache]
+            cache_provenance = None
+            cache_warnings = ""
```
tavinathanson: Doesn't seem like `cach_warnings` ever gets printed?

cohorts/load.py
jburos, line 1166: Yes - i copied the code from there. LMK if you want me to make this change

cohorts/load.py
```diff
@@ -1020,6 +1020,93 @@ def plot_joint(self, on, on_two=None, **kwargs):
...
+                    this_provenance = self.load_provenance(patient_cache_dir = patient_cache_dir)
+                    if not(cache_provenance):
+                        cache_provenance = this_provenance
+                    else:
+                        cache_warnings += _compare_provenance(this_provenance, cache_provenance)
+                if len(cache_warnings) == 0:
+                    provenance_summary[cache_name] = cache_provenance
+                else:
+                    provenance_summary[cache_name] = None
+        provenance_summary[u'dfhash'] = df_hash
```
tavinathanson: The treatment of `dfhash` feels a little weird to me. Here, it's treated as yet another cache name and below it's treated as yet another package, neither of which it is.

How about moving it out of `summarize_provenance` into `summarize_dataframe`?And that can be called from `summarize_unique_provenance`? Though, in the latter, how about an output like:

```
DataFrame hash: ___
Provenance: 
{cohorts: ...}
```

cohorts/load.py
```diff
@@ -1020,6 +1020,93 @@ def plot_joint(self, on, on_two=None, **kwargs):
         p = sb.jointplot(data=df, x=plot_cols[0], y=plot_cols[1])
         return p
 
+    def summarize_provenance(self):
```
tavinathanson: What do you think about calling this `cache_to_provenance` and the other one (gathered across caches) `summarize_provenance`; the `unique` vs. non-unique didn't really help me understand what was going on. Open to alternatives!

cohorts/load.py
```diff
@@ -1020,6 +1020,93 @@ def plot_joint(self, on, on_two=None, **kwargs):
...
+            `?cohorts.summarize_provenance` which optionally prints a single provenance dict if provenance is identical across all
+            cache_dirs.
+
+        """
+        provenance_summary = self.summarize_provenance()
+        df_hash = provenance_summary[u'dfhash']
+        first_provenance = None
+        cache_diff = ""
+        for cache in provenance_summary:
+            if cache != u'dfhash':
```
tavinathanson: As mentioned above, I find this a little weird.

tavinathanson:
@jburos Cool, LGTM % a few comments/suggestions!

cohorts/load.py
tavinathanson, line 1166: Yeah, don't think we should have duplicate code

cohorts/load.py
jburos: Sure - I like this approach. 

Conceptually, does `provenance` only refer to cached items? I wasn't sure how or where to put the data hash.  If it's not all provenance-related, perhaps this function should instead be renamed to something like `summarize_data_status` or `summarize_sources`. Curious to know what you think.  

Also is there a good way to separate how an object is printed from its structure?

cohorts/load.py
jburos: These get printed by `_compare_provenance()`. This object is just used to assess its length. Should _compare_provenance instead return a true/false?

cohorts/load.py
tavinathanson: I'm not too familiar with when/when not to use the term `provenance`; I imagine that it is appropriate to describe the hash. My concern is more around not treating the hash like a cache or a package. So maybe:

```
Provenance:

DataFrame hash: ___
Packages: 
{cohorts: ...}
```

I guess you could create a `Provenance` object that contains a package `dict` and a `df_hash`, and the `__str__` is the formatted string above?

cohorts/load.py
tavinathanson: Oh, I see. Yeah, I think returning a boolean there would make sense.

cohorts/load.py
jburos: That `load_from_cache` section has a customized warning that includes the patient id, which i'm not sure is necessary here .. i'll add that option this function so that we can reuse it.

cohorts/load.py
jburos, line 1166: Fixed in most recent version

cohorts/load.py
jburos: These two suggestions have been addressed in latest commits.

cohorts/load.py
jburos: OK I'm going to modify `_compare_provenance` to return the number of discrepancies, so that 0: no differences.

jburos:
Thanks @tavinathanson for all the feedback! 

Made a few changes based on your suggestions.

In brief:
1. separate dataframe_hash from provenance summary items
   - store this as an attribute in case it's expensive to compute
2. renamed functions for clarity, based on dict that is returned.  We now have 
   - `summarize_provenance_per_cache`: returns dict with provenance per cache_dir
   - `summarize_provenance`: summarizes provenance files across Cohort
   - `summarize_dataframe`: returns dataframe hash
   - `summarize_data_sources`: prints dataframe & provenance file summary
3. Clarified docstrings for summarize_provenance\* objects. 
   - Now state that they summarize contents of provenance files. These may or may not represent complete summary of data sources, but they are part of the picture.

output now looks like : 

```
> cohort.summarize_data_sources()
Out[5]:
{'dataframe_hash': -8604513298553306351,
 'provenance_file_summary': {u'cohorts': u'0+untagged.188.g6260531.dirty',
  u'isovar': u'0.0.6',
  u'mhctools': u'0.2.3',
  u'numpy': u'1.11.0',
  u'pandas': u'0.18.1',
  u'pyensembl': u'0.9.1',
  u'scipy': u'0.17.0',
  u'topiary': u'0.0.21',
  u'varcode': u'0.4.14'}}
```

Not sure if this warrants a separate Class of object (for a pretty `__str__` function), but this would be easy to add. For now I find the above sufficient.

tavinathanson:
Minor nits; feel free to ignore completely for now: would be good to stick to a single style for parameter/return docstrings. See some of the other functions e.g. https://github.com/hammerlab/cohorts/blob/master/cohorts/load.py#L478? Also, `discordant` vs. `discrepant` both used; choose one?

More importantly, Travis is failing because the isovar API is changed. Maybe change `git+git://github.com/hammerlab/isovar` to `isovar==0.0.6` and separately file an issue for us to get `cohorts` working with the latest?

tavinathanson:
Feel free to merge once Travis is green 👍

cohorts/load.py
```diff
@@ -1020,6 +1009,122 @@ def plot_joint(self, on, on_two=None, **kwargs):
...
+
+            == See also ==
+            
+            `?cohorts.Cohort.summarize_provenance_per_cache` which is used to summarize provenance for each existing cache_dir.
+
+        """
+        provenance_per_cache = self.summarize_provenance_per_cache()
+        summary_provenance = None
+        num_discordant = 0
+        for cache in provenance_per_cache:
```
tavinathanson, line 1106: Nit: `for cache, summary_provenance in provenance_per_cache.items()`?

tavinathanson, line 1106: Nevermind, I see why that's a problem here.

PR #75, opened 2016-07-10
Add responder_pfs_equals_os options

tavinathanson:
Sometimes we can't rely on the fact that a non-responder has PFS < OS and a responder has PFS == OS, so make this assumption optional.

Relevant discussion (sorry to link to a private repo): https://github.com/hammerlab/bladder-analyses/issues/82#issuecomment-231598825

@arahuja

### DIFF
--- a/cohorts/load.py
+++ b/cohorts/load.py
@@ -216,7 +216,7 @@ class Cohort(Collection):
                  extra_df_loaders=[],
                  join_with=None,
                  join_how="inner",
-                 responder_pfs_equals_os=True,
+                 responder_pfs_equals_os=False,
                  check_provenance=False,
                  polyphen_dump_path=None,
                  pageant_coverage_path=None,):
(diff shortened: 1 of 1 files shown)


arahuja:
👍   Do you want the default to be false? I like this as I've run into this with other sets as well. I'd be ok with dropping the constraint and just warning on load-in as necessary.

tavinathanson:
@arahuja I'm okay with defaulting to `False`, though have you specifically run into this constraint or also https://github.com/hammerlab/cohorts/pull/75/files#diff-481facac58e4fae6c10b3d93a889284eR260? I thought mostly you ran into not having PFS and OS values? Maybe another constraint called `require_pfs_and_os` that defaults to `False`, while this one can remain defaulted to `True`?

arahuja:
> Maybe another constraint called require_pfs_and_os that defaults to False, while this one can remain defaulted to True?  

Whichever makes sense, actually have run into both

tavinathanson:
@arahuja defaulted it to `False` and will defer the other constraint for another time. Good to merge?

arahuja:
👍

tavinathanson:
Oops, didn't wait for Travis green, my bad.

tavinathanson:
Oof, turns out I broke the build! That's what I get for being too cavalier. Fixing in https://github.com/hammerlab/cohorts/commit/e09195b928b3a28144d2be6ae5cf652d9fdb0257

PR #82, opened 2016-07-18
Fix effect counting

tavinathanson:
Since we're counting effects rather than variants generating effects, we might end up double-counting variants.

This does not appear to impact our bladder analyses.

@arahuja does this change make sense to you?

### DIFF
--- a/cohorts/functions.py
+++ b/cohorts/functions.py
@@ -43,7 +43,10 @@ def nonsynonymous_snv_count(row, cohort, filter_fn=None,
         filter_fn=filter_fn,
         **kwargs)
     if patient_id in patient_nonsynonymous_effects:
-        count = len(patient_nonsynonymous_effects[patient_id])
+        patient_variants = set()
+        for effect in patient_nonsynonymous_effects[patient_id]:
+            patient_variants.add(effect.variant)
+        count = len(patient_variants)
         if normalized_per_mb:
             count /= float(get_patient_to_mb(cohort)[patient_id])
         return count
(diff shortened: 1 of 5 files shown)


arahuja:
@tavinathanson for nonsyn=True don't we choose one effect per variant?

arahuja:
See: https://github.com/hammerlab/cohorts/blob/master/cohorts/load.py#L690

But this is true if we are counting all effects, I've run into that.

tavinathanson:
@arahuja I was scratching my head for a while, thinking that I must have been missing something! Indeed, I missed that we call `top_priority_effect_per_variant`! That probably explains why this change had no effect.

Given that the only counting functions I'm changing are those that use `top_priority_effect_per_variant`, this seems like a non-issue/close?

arahuja:
Yea, it does show that `load_effects` has some odd behavior. One way it gives all effects and another only nonsyn AND only per variant.

tavinathanson:
Perhaps a reasonable fix is to add `top_priority_effect_per_variant()` to all effects, as well as explain the non-bug in the counting functions for clarity?

arahuja:
> a reasonable fix is to add top_priority_effect_per_variant()

sounds good!

> as well as explain the non-bug in the counting functions for clarity

Maybe? Would be clearer? I think the assumption was always one effect per variant?

tavinathanson:
That must have been the assumption when writing those functions, though as a reader looking back at the code I was confused about why we didn't have a bug on our hands. Hence wanting to clarify :)

tavinathanson:
Yikes, was commiting to master by accident! Here's what I ended up with:

https://github.com/hammerlab/cohorts/commit/71e0082930b1c46dda5744ffdddfe85e19b279c0

Let me know if anything there doesn't look good and I'll make a new PR to fix it.

tavinathanson:
Note for anyone following this change involves re-generating the effects in the cache, though it shouldn't matter for most of our analyses (which were already taking the top priority effect).

Issue #69, opened 2016-07-07
builds against latest versions of isovar fails


jburos:
Currently, builds against the latest versions of isovar fail.   See, for example, [build #327](https://travis-ci.org/hammerlab/cohorts/builds/142872261). 

The error message says : 

```
ImportError: No module named protein_sequence
```

@tavinathanson mentions this is because the latest isovar API has changed.

For now, the `requirements.txt` file has been frozen at `isovar==0.0.6`.

jburos:
This issue cropped up again when referencing the latest version of `cohorts`: 

E.g. 

running `from cohorts.functions import *`

```
ImportErrorTraceback (most recent call last)
<ipython-input-1-c5376ccc9100> in <module>()
      9 import numpy as np
     10 import pandas as pd
---> 11 from cohorts.functions import *
     12 get_ipython().magic(u'matplotlib inline')

/mnt/ssd0/env/local/lib/python2.7/site-packages/cohorts/__init__.py in <module>()
     13 # limitations under the License.
     14 
---> 15 from .load import *
     16 
     17 from ._version import get_versions

/mnt/ssd0/env/local/lib/python2.7/site-packages/cohorts/load.py in <module>()
     35 from topiary import predict_epitopes_from_variants, epitopes_to_dataframe
     36 from topiary.sequence_helpers import contains_mutant_residues
---> 37 from isovar.protein_sequence import variants_to_protein_sequences_dataframe
     38 from pysam import AlignmentFile
     39 

ImportError: No module named protein_sequence
```

Here is the output when installing `cohorts` from git (note that `isovar` is not mentioned):

```
$ pip install git+git://github.com/hammerlab/cohorts --upgrade
Collecting git+git://github.com/hammerlab/cohorts
  Cloning git://github.com/hammerlab/cohorts to /tmp/pip-lrdXz6-build
Requirement already up-to-date: pandas>=0.15 in /mnt/ssd0/env/lib/python2.7/site-packages (from cohorts==0.1.1+103.g71e0082)
Requirement already up-to-date: seaborn>=0.7.0 in /mnt/ssd0/env/lib/python2.7/site-packages (from cohorts==0.1.1+103.g71e0082)
Requirement already up-to-date: scipy>=0.17.0 in /mnt/ssd0/env/lib/python2.7/site-packages (from cohorts==0.1.1+103.g71e0082)
[...]

PR #83, opened 2016-07-20
modify setup.py to reference requirements.txt

jburos:
I modified `setup.py` to load install_requires from `requirements.txt`. 

This should address situations like that seen in issue #69, which we thought we had fixed (by freezing `isovar` version) but it turned out we hadn't truly fixed -- we only changed isovar in `requirements.txt` but not in `setup.py`. Since requirements.ext was used to set up the travis testing environment, this failure to modify `setup.py` wasn't caught.

One downside of loading all packages from `requirements.txt` into `setup.py` is that we end up with utility packages like `pylint` included as requirements for cohorts. Same would go for packages required for testing, which may or may not be required for package function.

We could avoid this situation by having separate "package_requirements.txt" & "travis_requirements.txt" files, only one of which would be used to populate `setup.py`.  But, maintaining separate files introduces the potential for the problem to re-occur.  

Packages like `pylint` are hard-coded into the .travis-ci.yml and so could probably be taken out of this `requirements.txt` file entirely.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,11 +40,11 @@ except ImportError as e:
     pass
 
 # get the dependencies and installs
-with open(path.join(current_directory, 'requirements.txt'), encoding='utf-8') as f:
+with open(path.join(current_directory, "requirements.txt"), encoding='utf-8') as f:
     all_reqs = f.read().split('\n')
 
-install_requires = [x.strip() for x in all_reqs if 'git+' not in x]
-dependency_links = [x.strip().replace('git+','') for x in all_reqs if 'git+' in x]
+install_requires = [req.strip() for req in all_reqs if 'git+' not in req]
+dependency_links = [req.strip().replace('git+','') for req in all_reqs if 'git+' in req]
 
 if __name__ == "__main__":
     setup(
--- a/setup.py
+++ b/setup.py
@@ -14,9 +14,12 @@
 
 from __future__ import print_function
 import os
+from os import path
+from codecs import open
 from setuptools import setup
 import versioneer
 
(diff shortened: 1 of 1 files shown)


tavinathanson:
I'm not too worried about including `pylint`, `nose`, etc. We've even talked about always including the tests with the `pip install` as well. And it's not just for `Travis`, but anyone who wants to run tests/verify.

It might be cleaner to separate them out at some point, but for now this seems fine to me.

setup.py
```diff
@@ -14,9 +14,12 @@
 
 from __future__ import print_function
 import os
+from os import path
+from codecs import open
 from setuptools import setup
 import versioneer
 
+here = path.abspath(path.dirname(__file__))
```
tavinathanson: Duplicate logic/lines for grabbing the `README` path vs. the `requirements.txt`?

setup.py
```diff
@@ -37,6 +40,12 @@
     print("Failed to convert %s to reStructuredText", readme_filename)
     pass
 
+# get the dependencies and installs
+with open(path.join(here, 'requirements.txt'), encoding='utf-8') as f:
```
tavinathanson: Nit: double quotes to match the rest of the file?

setup.py
```diff
@@ -37,6 +40,12 @@
...
+    all_reqs = f.read().split('\n')
+
+install_requires = [x.strip() for x in all_reqs if 'git+' not in x]
```
tavinathanson: Style: `req` instead of `x` for readability?

tavinathanson:
LGTM, feel free to merge when Travis passes

Issue #80, opened 2016-07-15
Summarize provenance on init


jburos:

tavinathanson:
This was fixed in #84

PR #84, opened 2016-07-20
Report data sources

jburos:
Implemented #80 - summarize provenance on init. 

Prints a summary of data sources on initialization, so that analyses using the cohort data are earmarked with the version of data used for analysis.


tavinathanson:
LGTM! Merge away on Travis 📗

PR #85, opened 2016-07-21
first draft of clean_column_names utility function

jburos:
Added a function `strip_column_names` that converts column names to patsy-friendly strings.

**Note** the name has changed to `strip_column_names` (it was `clean_column_names`) in an attempt to be less confusing. Clean could be confused with presentation-ready column names.
## Example usage :

```
from cohorts.utils import strip_column_names
cohort = cohorts.init_cohort(...)
d = cohort.get_dataframe()
d = d.rename(columns = strip_column_names(d.columns))
```

You can now also do the above in one step, as:

```
d = cohort.get_dataframe(rename_cols=True)
```

  (this latter functionality is `off` by default, so the current behavior is unchanged)
## Behavior

In both cases, the function modifies column names as : 

```
1. convert meaningful punctuation to str (e.g. `<` becomes `lt`)
2. (optionally - not default) remove text within parens
3. convert remaining punctuation/whitespace -> _
4. change to lowercase
5. strip leading & trailing punctuation
```

It will additionally check for uniqueness of stripped column names, and will return original values if the revised names are non-unique.
## Examples

Some examples of how this maps from old to new names:

```
{'2-factor score': '2_factor_score',
 '5-factor score': '5_factor_score',
 'Age': 'age',
 'Albumin < 4': 'albumin_lt_4',
 'Alive Status': 'alive_status',
 'Baseline neutrophil to lymphocyte ratio': 'baseline_neutrophil_to_lymphocyte_ratio',
 'Best Response RECIST 1.1': 'best_response_recist_1_1',
 'Best response mRECIST': 'best_response_mrecist',
 'Date of Best Response': 'date_of_best_response',
 'ECOG >0': 'ecog_gt_0',
 'Hb < 10': 'hb_lt_10',
 'Liver Mets': 'liver_mets',
 'Number of Lines of Therapy': 'number_of_lines_of_therapy',
 'OS in days': 'os_in_days',
 'Ongoing Responder RECIST 1.1': 'ongoing_responder_recist_1_1',
 'PD-L1': 'pd_l1',
 'PFS (RECIST 1.1) in days': 'pfs_recist_1_1_in_days',
 'PFS (mRECIST 1.1) in days': 'pfs_mrecist_1_1_in_days',
 'Pack Years': 'pack_years',
 'Prior BCG': 'prior_bcg',
 'Response Mixed or Homogenous': 'response_mixed_or_homogenous',
 'Sequence Path': 'sequence_path',
```

### DIFF
--- a/cohorts/load.py
+++ b/cohorts/load.py
@@ -328,7 +328,8 @@ class Cohort(Collection):
         self.dataframe_hash = hash(str(df.sort_values("patient_id")))
         return df
 
-    def as_dataframe(self, on=None, col=None, join_with=None, join_how=None, **kwargs):
+    def as_dataframe(self, on=None, col=None, join_with=None, join_how=None,
+                    rename_cols=False, keep_paren_contents=True, **kwargs):
         """
         Return this Cohort as a DataFrame, and optionally include additional columns
         using `on`.
@@ -345,6 +346,18 @@ class Cohort(Collection):
 
         If `on` is a function or functions, kwargs is passed to those functions.
         Otherwise kwargs is ignored.
+
+        Other parameters
+        ----------------
+        `rename_cols`: (bool)
+                if True, then return columns using "clean" column names
+                ("clean" means lower-case names without punctuation other than `_`)
+                defaults to False
+        `keep_paren_contents`: (bool)
+                if True, then contents of column names within parens are kept.
+                if False, contents of column names within-parens are dropped.
+                Defaults to True
+        ----------
 
         Return : tuple or DataFrame
             <dataframe>
@@ -397,14 +410,13 @@ class Cohort(Collection):
             for key, value in on.iteritems():
                 col, df = apply_func(on=value, col=key, df=df)
                 cols.append(col)
-            return (cols, df)
         if type(on) == list:
             cols = []
             for i, elem in enumerate(on):
                 col = elem.__name__ if not is_lambda(elem) else "column_%d" % i
                 col, df = apply_func(on=elem, col=col, df=df)
                 cols.append(col)
-            return (cols, df)
+        return (cols, df)
 
     def load_dataframe(self, df_loader_name):
         """
--- a/cohorts/load.py
+++ b/cohorts/load.py
@@ -330,7 +330,7 @@ class Cohort(Collection):
         return df
 
     def as_dataframe(self, on=None, col=None, join_with=None, join_how=None,
-                    rename_cols=False, keep_paren_contents=True, **kwargs):
+                     rename_cols=False, keep_paren_contents=True, **kwargs):
         """
         Return this Cohort as a DataFrame, and optionally include additional columns
         using `on`.
@@ -351,14 +351,14 @@ class Cohort(Collection):
         Other parameters
         ----------------
         `rename_cols`: (bool)
-                if True, then return columns using "stripped" column names
-                ("stripped" means lower-case names without punctuation other than `_`)
-                See `utils.strip_column_names` for more details
-                defaults to False
+            if True, then return columns using "stripped" column names
+            ("stripped" means lower-case names without punctuation other than `_`)
+            See `utils.strip_column_names` for more details
+            defaults to False
         `keep_paren_contents`: (bool)
-                if True, then contents of column names within parens are kept.
-                if False, contents of column names within-parens are dropped.
-                Defaults to True
+            if True, then contents of column names within parens are kept.
+            if False, contents of column names within-parens are dropped.
+            Defaults to True
         ----------
 
         Return : tuple or DataFrame
@@ -420,8 +420,7 @@ class Cohort(Collection):
                 cols.append(col)
 
         if (rename_cols):
-            df = df.rename(columns=_strip_column_names(df.columns,
-                   keep_paren_contents=keep_paren_contents))
+            df = df.rename(columns=_strip_column_names(df.columns, keep_paren_contents=keep_paren_contents))
         return (cols, df)
 
     def load_dataframe(self, df_loader_name):
--- a/cohorts/load.py
+++ b/cohorts/load.py
@@ -38,6 +38,7 @@ from topiary.sequence_helpers import contains_mutant_residues
 from isovar.protein_sequence import variants_to_protein_sequences_dataframe
 from pysam import AlignmentFile
 
+from .utils import strip_column_names as _strip_column_names
 from .survival import plot_kmf
 from .plot import mann_whitney_plot, fishers_exact_plot, roc_curve_plot
 from .collection import Collection
@@ -350,8 +351,9 @@ class Cohort(Collection):
         Other parameters
         ----------------
         `rename_cols`: (bool)
-                if True, then return columns using "clean" column names
-                ("clean" means lower-case names without punctuation other than `_`)
+                if True, then return columns using "stripped" column names
+                ("stripped" means lower-case names without punctuation other than `_`)
+                See `utils.strip_column_names` for more details
                 defaults to False
         `keep_paren_contents`: (bool)
                 if True, then contents of column names within parens are kept.
@@ -416,6 +418,10 @@ class Cohort(Collection):
                 col = elem.__name__ if not is_lambda(elem) else "column_%d" % i
                 col, df = apply_func(on=elem, col=col, df=df)
                 cols.append(col)
+
+        if (rename_cols):
+            df = df.rename(columns=_strip_column_names(d.columns,
+                   keep_paren_contents=keep_paren_contents))
         return (cols, df)
 
     def load_dataframe(self, df_loader_name):
--- a/cohorts/load.py
+++ b/cohorts/load.py
@@ -420,7 +420,7 @@ class Cohort(Collection):
                 cols.append(col)
 
         if (rename_cols):
-            df = df.rename(columns=_strip_column_names(d.columns,
+            df = df.rename(columns=_strip_column_names(df.columns,
                    keep_paren_contents=keep_paren_contents))
         return (cols, df)
 
--- a/cohorts/load.py
+++ b/cohorts/load.py
@@ -420,7 +420,9 @@ class Cohort(Collection):
                 cols.append(col)
 
         if (rename_cols):
-            df = df.rename(columns=_strip_column_names(df.columns, keep_paren_contents=keep_paren_contents))
+            rename_dict = _strip_column_names(df.columns, keep_paren_contents=keep_paren_contents)
+            df.rename(columns=rename_dict, inplace=True)
+            cols = [rename_dict[col] for col in cols]
         return (cols, df)
 
     def load_dataframe(self, df_loader_name):
--- a/cohorts/utils.py
+++ b/cohorts/utils.py
@@ -16,10 +16,10 @@ import re
 import logging
 
 
-def _clean_column_name(col_name, keep_paren_contents=True):
+def _strip_column_name(col_name, keep_paren_contents=True):
     """
     Utility script applying several regexs to a string.
-    Intended to be used by `clean_column_names`.
+    Intended to be used by `strip_column_names`.
 
     This function will:
         1. replace informative punctuation components with text
@@ -41,7 +41,7 @@ def _clean_column_name(col_name, keep_paren_contents=True):
 
     Examples
     --------
-    > print([_clean_column_name(col) for col in ['PD-L1','PD L1','PD L1_']])
+    > print([_strip_column_name(col) for col in ['PD-L1','PD L1','PD L1_']])
     """
     # start with input
     new_col_name = col_name
@@ -77,12 +77,13 @@ def _clean_column_name(col_name, keep_paren_contents=True):
     return new_col_name.lower()
 
 
-def clean_column_names(cols, keep_paren_contents=True):
+def strip_column_names(cols, keep_paren_contents=True):
     """
     Utility script for renaming pandas columns to patsy-friendly names.
-    Revised names have:
-        - punctuation and whitespace -> text or _
-        - all text in lower case
+
+    Revised names have been:
+        - stripped of all punctuation and whitespace (converted to text or `_`)
+        - converted to lower case
 
     Takes a list of column names, returns a dict mapping
     names to revised names.
@@ -113,21 +114,22 @@ def clean_column_names(cols, keep_paren_contents=True):
       'PD L1 (>1)': pd.Series([0., 1., 1., 0.], index=['a', 'b', 'c', 'd']),
       }
     > df = pd.DataFrame(df)
-    > df = df.rename(columns = clean_column_names(df.columns))
+    > df = df.rename(columns = strip_column_names(df.columns))
 
     ## observe, by comparison
-    > df2 = df.rename(columns = clean_column_names(df.columns,
+    > df2 = df.rename(columns = strip_column_names(df.columns,
         keep_paren_contents=False))
     """
     logger = logging.getLogger()
 
-    # clean/replace punctuation
+    # strip/replace punctuation
     new_cols = [
-        _clean_column_name(col, keep_paren_contents=keep_paren_contents)
+        _strip_column_name(col, keep_paren_contents=keep_paren_contents)
         for col in cols]
 
     if not(len(new_cols) == len(set(new_cols))):
-        logger.warning('Warning: clean_column_names will introduce duplicate names. Reverting column names to the original.')
+        logger.warning('Warning: strip_column_names (if run) would introduce duplicate names.'
+            ' Reverting column names to the original.')
         return dict(zip(cols, cols))
 
     return dict(zip(cols, new_cols))
--- a/cohorts/utils.py
+++ b/cohorts/utils.py
@@ -38,7 +38,7 @@ def _clean_column_name(x, keep_parens = True):
 
         Examples
         --------- 
-        > print([clean_column_name(col) for col in ['PD-L1','PD L1','PD L1_']])
+        > print([_clean_column_name(col) for col in ['PD-L1','PD L1','PD L1_']])
 
     """
     ## start with input
--- a/cohorts/utils.py
+++ b/cohorts/utils.py
@@ -14,6 +14,7 @@
 
 import re
 import logging
+import warnings
 
 
 def _strip_column_name(col_name, keep_paren_contents=True):
@@ -128,8 +129,11 @@ def strip_column_names(cols, keep_paren_contents=True):
         for col in cols]
 
     if not(len(new_cols) == len(set(new_cols))):
-        logger.warning('Warning: strip_column_names (if run) would introduce duplicate names.'
-                       ' Reverting column names to the original.')
+        warn_str = 'Warning: strip_column_names (if run) would introduce duplicate names.'
+        warn_str += ' Reverting column names to the original.'
+
+        logger.warning(warn_str)
+        warnings.warn(warn_str)
         print('Warning: strip_column_names would introduce duplicate names. Please fix & try again.')
         return dict(zip(cols, cols))
 
--- a/cohorts/utils.py
+++ b/cohorts/utils.py
@@ -128,7 +128,7 @@ def strip_column_names(cols, keep_paren_contents=True):
         _strip_column_name(col, keep_paren_contents=keep_paren_contents)
         for col in cols]
 
-    if not(len(new_cols) == len(set(new_cols))):
+    if len(new_cols) != len(set(new_cols)):
         warn_str = 'Warning: strip_column_names (if run) would introduce duplicate names.'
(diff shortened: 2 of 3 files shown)


cohorts/utils.py
```diff
@@ -0,0 +1,124 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import re 
+
+def _clean_column_name(x, keep_parens = True):
+    """
```
tavinathanson, line 21: Style: lines should be indented at the same level as the `"""`: https://www.python.org/dev/peps/pep-0008/#documentation-strings

cohorts/utils.py
```diff
@@ -0,0 +1,124 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import re 
+
+def _clean_column_name(x, keep_parens = True):
```
tavinathanson: Better name than `x`?

cohorts/utils.py
```diff
@@ -0,0 +1,124 @@
...
+
+        This function will: 
+            1. replace informative punctuation components with text
+            2. (optionally) remove text within parentheses
+            3. replace remaining punctuation/whitespace with _
+            4. remove trailing punctuation/whitespace
+
+        Parameters 
+        -----------
+        str (str): input character string
```
tavinathanson: `str` vs. `x`

cohorts/utils.py
```diff
@@ -0,0 +1,124 @@
...
+        keep_parens (logical): control behavior of within-paren elements of text
```
tavinathanson: Maybe `keep_paren_contents` would be more accurate?

cohorts/utils.py
```diff
@@ -0,0 +1,124 @@
...
+             - if True, (the default) all text within parens retained
+             - if False, text within parens will be removed from the field name
+
+        Returns
+        --------
+        modified string for new field name 
+
+
```
tavinathanson, line 62: Unnecessary newline

cohorts/utils.py
```diff
@@ -0,0 +1,124 @@
...
+    punct_replacement = '_'
+    newx = re.sub(punct_pattern, punct_replacement, newx)
+    
+    ## remove trailing _ if it exists (if last char was punctuation)
+    newx = re.sub('_$', '', newx)
+    return newx
+
+def clean_column_names(cols, keep_parens = True):
+    """
+        Utility script for renaming pandas columns to patsy-friendly names.
```
tavinathanson: Same comment re indentation

test/test_utils.py
```diff
@@ -0,0 +1,39 @@
...
+      }
+    d = pd.DataFrame(d)
+    ## should not error & should rename columns
+    d2 = d.rename(columns = clean_column_names(d.columns))
+    ok_((d2.columns != d.columns).any())
+
+    ## should not rename columns -- should error
+    d3 = d.rename(columns = clean_column_names(d.columns, keep_parens = False))
+    ok_((d3.columns == d.columns).all())
+
```
tavinathanson, line 50: Excess newlines

test/test_utils.py
```diff
@@ -0,0 +1,39 @@
...
+    res = [_clean_column_name(col) for col in ['PD-L1','PD L1','PD L1_']]
+    eq_(res, ['PD_L1','PD_L1','PD_L1'])
+
+def test_column_names():
+    d = {'one' : pd.Series([1., 2., 3.], index=['a', 'b', 'c']),
+      'two' : pd.Series([1., 2., 3., 4.], index=['a', 'b', 'c', 'd']),
+      'PD L1 (value)': pd.Series([1., 2., 3., 4.], index=['a', 'b', 'c', 'd']),
+      'PD L1 (>1)': pd.Series([0., 1., 1., 0.], index=['a', 'b', 'c', 'd']),
+      }
+    d = pd.DataFrame(d)
```
tavinathanson: Very minor: might be good to use `df` vs. `d`, seems to be the standard way to name a `DataFrame` e.g. http://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.html

cohorts/utils.py
```diff
@@ -0,0 +1,124 @@
...
+        ## observe, by comparison
+        > d2 = d.rename(columns = clean_column_names(d.columns, keep_parens = False))
+
+    """
+    ## clean/replace punctuation
+    newcols = [_clean_column_name(col, keep_parens = keep_parens) for col in cols]
+    
+    ## confirm no duplicates in remaining columns
+    if not(len(newcols) == len(cols)):
+        print('Warning: length of newcols not the same as cols')
```
tavinathanson: Worth raising an error here?

cohorts/utils.py
```diff
@@ -0,0 +1,124 @@
...
+        --------- 
+        > print([_clean_column_name(col) for col in ['PD-L1','PD L1','PD L1_']])
+
+    """
+    ## start with input
+    newx = x
+    
+    ## replace meaningful punctuation with text equivalents
+    ## surround each with whitespace to enforce consistent use of _
+    newx = re.sub('<=', ' le ', newx)
```
tavinathanson: Maybe avoid the repeated `re.sub` e.g.

```
punctuation_to_text = {"=<": "le", ...}
for punctuation, punctuation_text in punctuation_to_text.items():
    col_name = col_name.replace(punctuation, punctuation_text)
```

cohorts/utils.py
```diff
@@ -0,0 +1,124 @@
...
+    if not(keep_parens):
+        newx = re.sub('\([^)]*\)', '', newx)
+    
+    ## replace remaining punctuation/whitespace with _
+    punct_pattern = '[\W_]+'
+    punct_replacement = '_'
+    newx = re.sub(punct_pattern, punct_replacement, newx)
+    
+    ## remove trailing _ if it exists (if last char was punctuation)
+    newx = re.sub('_$', '', newx)
```
tavinathanson: Maybe `newx.strip("_")` is cleaner?

cohorts/utils.py
```diff
@@ -0,0 +1,124 @@
...
+            - punctuation and whitespace -> text or _
+            - all text in lower case 
+
+        Takes a list of column names, returns a dictionary mapping names to revised names.
+
+        If there are any concerns with the conversion, this will print a warning & return original column names. 
+        
+        Parameters 
+        --------------
+            cols (list): list of strings containing column names
```
tavinathanson: I think this is actually two indentations too far in: should be lined up with `Parameters`, which should be lined up with `"""`

cohorts/utils.py
```diff
@@ -0,0 +1,124 @@
...
+        
+        ## observe, by comparison
+        > d2 = d.rename(columns = clean_column_names(d.columns, keep_parens = False))
+
+    """
+    ## clean/replace punctuation
+    newcols = [_clean_column_name(col, keep_parens = keep_parens) for col in cols]
+    
+    ## confirm no duplicates in remaining columns
+    if not(len(newcols) == len(cols)):
```
tavinathanson: When would this be not true? Versus `len(set(newcols))`?

cohorts/utils.py
```diff
@@ -0,0 +1,124 @@
...
+              }
+        > d = pd.DataFrame(d)
+        > d = d.rename(columns = clean_column_names(d.columns))
+        
+        ## observe, by comparison
+        > d2 = d.rename(columns = clean_column_names(d.columns, keep_parens = False))
+
+    """
+    ## clean/replace punctuation
+    newcols = [_clean_column_name(col, keep_parens = keep_parens) for col in cols]
```
tavinathanson: Minor style: `new_cols` vs. `newcols`

cohorts/utils.py
```diff
@@ -0,0 +1,124 @@
...
+    
+    ## confirm no duplicates in remaining columns
+    if not(len(newcols) == len(cols)):
+        print('Warning: length of newcols not the same as cols')
+        return dict(zip(cols, cols))
+    
+    if not(len(newcols) == len(set(newcols))):
```
tavinathanson: Why is `lower`ing not just a part of `_clean_column_name`?

cohorts/utils.py
```diff
@@ -0,0 +1,124 @@
...
+        print('Warning: cleanup introduces duplicate names. Please resolve & try again.')
+        return dict(zip(cols, cols))
+    
+    ## modify all column names to lower case
+    newcols_lc = [col.lower() for col in newcols]
```
tavinathanson: Why is `lower`ing not just a part of `_clean_column_name`?

tavinathanson:
LGTM! % minor style comments

cohorts/utils.py
jburos: This is a sanity check - one checks for uniqueness the other checks whether the list-comprehension worked properly

tavinathanson: Two thoughts: `assert` is good for when we want to test basic things that shouldn't actually go wrong. That said, I don't think we need to take up space testing that Python list comprehensions work?

cohorts/utils.py
jburos: A few reasons:
1. So this can be turned off without having to pass parameter to `_clean_column_names`
2. Separating steps would, I thought, be somewhat cleaner.
3. So uniqueness could be checked before & after case-conversion. Ideally I want to have the non-uniqueness message be more helpful, like "after step X these columns were equivalent - please fix & try again". & actually list the duplicate columns.

cohorts/utils.py
jburos: Would this also strip leading _ ?

cohorts/utils.py
jburos: I like this - then the punct-to-string can be more easily extended. Thanks!

tavinathanson:
@jburos also just occurred to me: why manual `d.rename(columns = clean_column_names(d.columns))` vs. `cohort.as_dataframe(rename_cols=True, keep_paren_contents=True)`, with `True` by default?

cohorts/utils.py
tavinathanson: It would, is that a problem?

cohorts/utils.py
jburos: Perhaps .. was thinking about how to make this more friendly for interactive use. Is there a way to raise a warning?

tavinathanson: I think the standard warning protocol is via `logging`, which I assume is generally preferable over `print` (and `cohorts` could do better at that as a whole). So I guess the question here is whether the issue is worth halting completely over. I'll defer to you on that call?

tavinathanson: Oops, I think `warnings.warn("warning message", TypeOfWarning)` might be even better? I don't know too much about this, sadly!

jburos:
@tavinathanson I would suggest we _also_ support that usage. 

I would recommend we keep the functional use separate since this utility 
1. could be used outside of cohorts
2. can be applied to a subset of column names easily this way, and 
3. including it in the as_data_frame call could make errors converting columns (ie string handling resulting in non-unique column names) difficult to address

But, I do agree that this additional functionality could be helpful -- e.g. when you want to apply coxph model to outcome vars & this is only exposed via the `Cohort` object.

Curious to know what you think.

cohorts/utils.py
tavinathanson: Ignore everything I said! I guess `warning.warn` would raise an exception depending on the current settings for warning suppression.

tavinathanson:
@jburos I agree that this should be accessible outside `as_dataframe`, but `as_dataframe` seems like a great place to use it. Not too worried about errors in the conversion; if we clearly `warn` when they come up, that seems fine?

cohorts/utils.py
tavinathanson: I think checking uniqueness at every step is probably over-engineering this; if a uniqueness error occurs, I think it should be enough to spit out which columns are duplicates, regardless of what step made them so?

I think, readability-wise, it's more intuitive to have the `_clean_column_name` function do everything to each individual column name, while the `clean_column_names` function iterates over them. That would be my assumption, and any deviation seems like added confusion?

Not clear what 1 means: how do you turn it off?

cohorts/utils.py
jburos: Maybe - in some cases, I have used a leading underscore to include information about the version of data in that row. But otherwise I agree with you it would _usually_ be good to strip this.

I will implement strip for now & we'll cross the leading-underscore problem if/when we encounter it

cohorts/utils.py
jburos: True. I guess I was thinking if there was an None value or something , but this probably wouldn't catch that anyways.

cohorts/utils.py
jburos: I haven't added that parameter, but conceivably a user might not want to convert case

jburos:
@tavinathanson I modified this in two places : 
1. Renamed function (& helper function) to `strip_column_names`.
2. Added option for `rename_cols` to `as_dataframe` (defaults to False).

Would you mind reviewing these two changes again? 

Thanks & apologies for the CR-overload...

cohorts/load.py
```diff
@@ -397,14 +412,17 @@ def is_lambda(func):
...
             cols = []
             for i, elem in enumerate(on):
                 col = elem.__name__ if not is_lambda(elem) else "column_%d" % i
                 col, df = apply_func(on=elem, col=col, df=df)
                 cols.append(col)
-            return (cols, df)
+
+        if (rename_cols):
+            df = df.rename(columns=_strip_column_names(d.columns,
+                   keep_paren_contents=keep_paren_contents))
```
jburos: @tavinathanson wasn't sure here what I should do with `cols` - welcome your advice on this part

cohorts/load.py
```diff
@@ -328,7 +329,8 @@ def _as_dataframe_unmodified(self, join_with = None, join_how = None):
         self.dataframe_hash = hash(str(df.sort_values("patient_id")))
         return df
 
-    def as_dataframe(self, on=None, col=None, join_with=None, join_how=None, **kwargs):
+    def as_dataframe(self, on=None, col=None, join_with=None, join_how=None,
+                    rename_cols=False, keep_paren_contents=True, **kwargs):
```
tavinathanson: Style nit: `rename_cols` should be directly underneath the paren

cohorts/load.py
```diff
@@ -346,6 +348,19 @@ def as_dataframe(self, on=None, col=None, join_with=None, join_how=None, **kwarg
         If `on` is a function or functions, kwargs is passed to those functions.
         Otherwise kwargs is ignored.
 
+        Other parameters
+        ----------------
+        `rename_cols`: (bool)
+                if True, then return columns using "stripped" column names
```
tavinathanson: Indent by 4 spaces vs. 8? https://www.python.org/dev/peps/pep-0008/#indentation

cohorts/load.py
```diff
@@ -397,14 +412,17 @@ def is_lambda(func):
...
         if type(on) == list:
             cols = []
             for i, elem in enumerate(on):
                 col = elem.__name__ if not is_lambda(elem) else "column_%d" % i
                 col, df = apply_func(on=elem, col=col, df=df)
                 cols.append(col)
-            return (cols, df)
+
+        if (rename_cols):
+            df = df.rename(columns=_strip_column_names(df.columns,
```
tavinathanson: I apologize for the confusing nature of this code, but I think this will break in some instances when the name of the function, e.g. `missense_snv_counT` gets stripped here, however it doesn't get stripped in the `cols` variable.

Maybe, instead of putting this `if (rename_cols)` here, you should put it inside `apply_func` above, replacing `df[col] = df.apply(func, axis=1)`?

Something to replace both the column name and the dataframe e.g.:

```
applied = df.apply(func, axis=1)
if (rename_cols):
    df.rename(...)
    col = _strip_column_name(...)
df[col] = applied
```

cohorts/load.py
```diff
@@ -397,14 +412,16 @@ def is_lambda(func):
...
         if type(on) == list:
             cols = []
             for i, elem in enumerate(on):
                 col = elem.__name__ if not is_lambda(elem) else "column_%d" % i
                 col, df = apply_func(on=elem, col=col, df=df)
                 cols.append(col)
-            return (cols, df)
+
+        if (rename_cols):
+            df = df.rename(columns=_strip_column_names(df.columns, keep_paren_contents=keep_paren_contents))
```
tavinathanson: Sorry for the confusing nature of this code! I think the `DataFrame` itself might get column-renamed while the `cols` variable won't, which could lead to issues.

I think `cols = _strip_column_names(cols, keep_paren_contents=keep_paren_contents)` would address that?

tavinathanson: Also, you can do an in-place rename with `df.rename(inplace=True)` to avoid `df = df.rename`

cohorts/utils.py
```diff
@@ -0,0 +1,136 @@
...
+
+    # replace remaining punctuation/whitespace with _
+    punct_pattern = '[\W_]+'
+    punct_replacement = '_'
+    new_col_name = re.sub(punct_pattern, punct_replacement, new_col_name)
+
+    # remove leading/trailing _ if it exists (if last char was punctuation)
+    new_col_name = new_col_name.strip("_")
+
+    # TODO: check for empty string
```
tavinathanson, line 75: Not sure if this is intentional, but seems fine to leave the `TODO` if you make an issue for it

cohorts/utils.py
```diff
@@ -0,0 +1,136 @@
...
+    cols (list): list of strings containing column names
+    keep_paren_contents (logical):
+        controls behavior of within-paren elements of text
+         - if True, (the default) all text within parens retained
+         - if False, text within parens will be removed from the field name
+
+    Returns
+    -------
+
+    dict mapping cols -> newcols
```
tavinathanson: `new_col`

cohorts/utils.py
```diff
@@ -0,0 +1,136 @@
...
+
+    # strip/replace punctuation
+    new_cols = [
+        _strip_column_name(col, keep_paren_contents=keep_paren_contents)
+        for col in cols]
+
+    if not(len(new_cols) == len(set(new_cols))):
+        logger.warning('Warning: strip_column_names (if run) would introduce duplicate names.'
+                       ' Reverting column names to the original.')
+        print('Warning: strip_column_names would introduce duplicate names. Please fix & try again.')
```
tavinathanson, line 137: Is the reason you're `print`ing in addition to logging because it's not showing up in the notebook? I was reading more of this to try to understand how this should work: http://stackoverflow.com/questions/18786912/get-output-from-the-logging-module-in-ipython-notebook; but I admit that I'm fairly naive and this is fine for now until we get more clarity.

cohorts/utils.py
```diff
@@ -0,0 +1,136 @@
...
+        keep_paren_contents=False))
+    """
+    logger = logging.getLogger()
+
+    # strip/replace punctuation
+    new_cols = [
+        _strip_column_name(col, keep_paren_contents=keep_paren_contents)
+        for col in cols]
+
+    if not(len(new_cols) == len(set(new_cols))):
```
tavinathanson: `len(new_cols) != len(set(new_cols))`?

cohorts/load.py
tavinathanson: Have a simpler thought elsewhere; please ignore this comment!

tavinathanson:
@jburos looks good! Once you address the remaining comments, feel free to merge without another review pass (though happy to give more feedback if useful)

cohorts/load.py
jburos: Thanks - was wondering what to do about `cols` -- this will be a bit tricky to do cleanly since I won't know at this point if the column-rename was successful... feel like I should look these up in the dictionary returned by the original _strip_column_names on the df. 

I'll try to write up a test case for this.

tavinathanson: I can't think of a scenario where the rename would fail on a column name in the `DataFrame` that it would not fail on in the `cols`? I agree that this feels a little hacky, and if you want to do it more cleanly feel free; fine with me either way.

cohorts/utils.py
jburos, line 137: Yes - that's why. I also ended up adding a `warning.warn` since the `logger.warning` message didn't show up in either the console or the ipynb -- appears this takes a separate setup which I was reading about here : https://docs.python.org/2.6/library/logging.html#configuring-logging

Definitely doesn't belong in this utils module, though should probably be addressed at some point .. for now I have found duplicating these messages helpful.

PR #86, opened 2016-07-22
Make provenance printing optional

tavinathanson:
Also fix some docstrings.

@jburos

### DIFF
--- a/cohorts/load.py
+++ b/cohorts/load.py
@@ -216,7 +216,7 @@ class Cohort(Collection):
     check_provenance : bool
         Verify that the cached provenance is equal to the current environment.
     print_provenance : bool
-        Print a cached provenance summary.
+        Print a summary of cache file provenance.
     polyphen_dump_path : str
         Path to a Polyphen database dump.
     pageant_coverage_path : str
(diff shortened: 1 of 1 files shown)


cohorts/load.py
```diff
@@ -213,6 +213,14 @@ class Cohort(Collection):
     responder_pfs_equals_os : bool
         Ensure that the PFS values for responders (not progressed) are equal to
         OS values.
+    check_provenance : bool
+        Verify that the cached provenance is equal to the current environment.
+    print_provenance : bool
+        Print a cached provenance summary.
```
jburos: Technically, this is a summary of the cache provenance .. not a cached summary of the provenance files. 
Maybe say "Print a summary of cache file provenance"?

jburos:
Thanks @tavinathanson - just one minor suggestion, but otherwise looks good to me. Thanks for updating the docstrings

cohorts/load.py
tavinathanson: Was supposed to be read as (cached provenance) summary vs. cached (provenance summary), but I'll update :)

jburos:
One comment - this makes me want to know, in which scenarios you'd like to turn off the data source summary? 

Please LMK (or maybe create issues) if there are modifications that should be made to that process (ie to check but not print, or to print but not warn, to exclude certain packages from the summary, etc).

tavinathanson:
I have notebooks that initialize `Cohort`s inside a loop, and I don't want so many summaries printed. Open to smarter ways of handling this, though this seemed like a simple bandaid for starters. Will make an issue!

PR #90, opened 2016-07-22
Add filter_fn to Cohort

tavinathanson:
Add `Cohort` defaults for `filter_fn`, `normalized_per_mb`, and Pageant depths.

@arahuja

### DIFF
--- a/cohorts/functions.py
+++ b/cohorts/functions.py
@@ -95,15 +95,16 @@ def neoantigen_count(row, cohort, filter_fn=None,
         return count
     return np.nan
 
-def expressed_missense_snv_count(row, cohort, filter_fn=None, **kwargs):
+def expressed_missense_snv_count(row, cohort, **kwargs):
+    filter_fn = kwargs.pop("filter_fn")
     def expressed_filter_fn(filterable_effect):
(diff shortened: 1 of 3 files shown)


arahuja:
Cool, I think having the cohort have a fallback is great.

A few thinking points, feel free to move ahead though if you don't agree:
- `min_coverage_normal_depth` is only used for the ensembl part right and unlinked from the filter functions, could that be confusing?
- We definitely need a way to not filter which `no_fitler` solves, but for some reason, I'd rather `None` did that? Does it make sense for `filter_fn` to have 3 possible args: `'default'`, a filter fn, or None. Where `'default'` goes to the cohort? This way the user never needs to know of anything special, either the default happens or the user specifies a filter function or `None`? I worry about the discoverability of `no_filter` and just `None` being more natural?

Just a few thoughts, but looks good!

tavinathanson:
@arahuja all good points!

Re `min_coverage_normal_depth`, I agree that it's confusing but I'm not sure of the best way to keep it more clear?

Re `None` vs. `no_filter`, I like your idea. I'll add an issue to fix that so that this code can be merged quickly and used.

PR #95, opened 2016-07-25
Fix expressed_missense_snv_count for None filter_fn

tavinathanson:
`expressed_missense_snv_count` is currently broken when no `filter_fn` arg is present. Fixes that.

@arahuja 

FYI @jburos

### DIFF
--- a/cohorts/functions.py
+++ b/cohorts/functions.py
@@ -96,7 +96,7 @@ def neoantigen_count(row, cohort, filter_fn=None,
     return np.nan
 
 def expressed_missense_snv_count(row, cohort, **kwargs):
-    filter_fn = kwargs.pop("filter_fn")
+    filter_fn = kwargs.pop("filter_fn", None)
     def expressed_filter_fn(filterable_effect):
         if filter_fn is not None:
             return filter_fn(filterable_effect) and effect_expressed_filter(filterable_effect)
--- a/cohorts/functions.py
+++ b/cohorts/functions.py
@@ -95,13 +95,16 @@ def neoantigen_count(row, cohort, filter_fn=None,
         return count
     return np.nan
 
-def expressed_missense_snv_count(row, cohort, **kwargs):
-    filter_fn = kwargs.pop("filter_fn", None)
+def expressed_missense_snv_count(row, cohort, filter_fn=None,
+                                 normalized_per_mb=None):
+    filter_fn = first_not_none_param([filter_fn, cohort.filter_fn], no_filter)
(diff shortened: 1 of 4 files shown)


cohorts/functions.py
```diff
@@ -96,7 +96,7 @@ def neoantigen_count(row, cohort, filter_fn=None,
     return np.nan
 
 def expressed_missense_snv_count(row, cohort, **kwargs):
-    filter_fn = kwargs.pop("filter_fn")
+    filter_fn = kwargs.pop("filter_fn", None)
```
arahuja: Just checking this `None` means use `Cohort` default, correct?

tavinathanson: Yup, I'm fairly certain that's correct. Did some double checking the other day, but should really have some tests.

arahuja:
👍

cohorts/functions.py
tavinathanson: @arahuja glad you asked, since this was wrong! Felt a little unsure so wrote a test that revealed the problem. Thanks for pointing it out.

tavinathanson: (I forgot that `expressed_filter_fn` is the new `filter_fn`, which doesn't give `Cohort.filter_fn` a chance to be used; since by the time we check for the default, we think we already have a `filter_fn`: `expressed_filter_fn`.)

tavinathanson:
@jburos FYI `expressed_missense_snv_count` had a bug that wasn't properly fixed until now, in https://github.com/hammerlab/cohorts/pull/95/commits/b9d625b8bd066ef27c075535fe45230d7733fe22. Whoops 👎

arahuja:
@tavinathanson I think I've lost track a bit of how things work here, so I'll take a look at this in more detail

cohorts/functions.py
```diff
@@ -95,13 +95,16 @@ def neoantigen_count(row, cohort, filter_fn=None,
...
-def expressed_missense_snv_count(row, cohort, **kwargs):
-    filter_fn = kwargs.pop("filter_fn")
+def expressed_missense_snv_count(row, cohort, filter_fn=None,
+                                 normalized_per_mb=None):
+    filter_fn = first_not_none_param([filter_fn, cohort.filter_fn], no_filter)
+    normalized_per_mb = first_not_none_param([normalized_per_mb, cohort.normalized_per_mb], False)
     def expressed_filter_fn(filterable_effect):
         if filter_fn is not None:
             return filter_fn(filterable_effect) and effect_expressed_filter(filterable_effect)
         return effect_expressed_filter(filterable_effect)
```
tavinathanson: @arahuja pointed out that this line should never be reached.

PR #98, opened 2016-07-28
add load_kallisto()

e5c:
Enables,
`df_kallisto = cohort.as_dataframe(join_with="kallisto")`


cohorts/load.py
```diff
@@ -764,6 +767,43 @@ def _load_single_patient_effects(self, patient, only_nonsynonymous, filter_fn):
             patient=patient,
             filter_fn=filter_fn)
 
+    def load_kallisto(self):
+        """
+        Load a Kallisto gene expression data for a cohort
+
+        Parameters
```
tavinathanson, line 780: Is this a good practice to have an empty parameter list result in an empty docstring parameter list? (Not sure, just asking.)

cohorts/load.py
```diff
@@ -764,6 +767,43 @@ def _load_single_patient_effects(self, patient, only_nonsynonymous, filter_fn):
             patient=patient,
             filter_fn=filter_fn)
 
+    def load_kallisto(self):
+        """
+        Load a Kallisto gene expression data for a cohort
```
arahuja: technically this is loading transcript quantification. 

It might be worthwhile to annotate the gene's here as well, though it's obvious how to aggregate to gene level. Summing should be fine, but [`tximport`](https://github.com/mikelove/tximport) tries some other strategies as well.

tavinathanson:
LGTM!

I wonder if we can simplify some of the functions that are glorified `read_csv`s so that we don't need two functions per?

Okay with me to defer that for later, though.

tavinathanson:
Perhaps my comment https://github.com/hammerlab/cohorts/pull/98#issuecomment-236011834 isn't applicable in light of @arahuja's comment (https://github.com/hammerlab/cohorts/pull/98#discussion_r72693593), since it sounds like this can be more than a glorified `read_csv`.

arahuja:
LGTM!

cohorts/load.py
```diff
@@ -764,6 +773,50 @@ def _load_single_patient_effects(self, patient, only_nonsynonymous, filter_fn):
...
+            copy=False
+        )
+
+        ensembl_release = cached_release(self.ensembl_version)
+
+        kallisto_data['gene_name'] = \
+            kallisto_data['target_id'].map(lambda t: ensembl_release.gene_name_of_transcript_id(t))
+
+        return kallisto_data
+
```
tavinathanson, line 803: Style nit: remove this newline?

tavinathanson:
@e5c can you call it `kallisto_ensembl_version` to make it more obvious where it's used? At first glance I'd assume that it's also used for interpreting VCFs. (But there we're inferring it from the VCF.) If it becomes broadly used we can rename it then to something more general?

cohorts/load.py
```diff
@@ -239,6 +245,7 @@ class Cohort(Collection):
     def __init__(self,
                  patients,
                  cache_dir,
+                 ensembl_version,
```
tavinathanson: Since Kallisto isn't required when creating a `Cohort`, can we default this to something? Maybe default it to `None` and then raise a `ValueError` in `load_kallisto` if the arg is `None`?

test/test_basic.py
```diff
@@ -51,6 +51,7 @@ def make_simple_cohort(merge_type="union", **kwargs):
 
     return Cohort(
         patients=patients,
+        ensembl_version=75,
```
tavinathanson: See comment https://github.com/hammerlab/cohorts/pull/98/files#r72807318; I think it's cleaner to keep the tests the same and not require this param when Kallisto isn't relevant?

tavinathanson:
LGTM!

PR #99, opened 2016-07-31
Fix varcode at 0.4.15 until API updates

arahuja:
Need to figure out the necessary API updates, but until then this should fix master


e5c:
lgtm

PR #103, opened 2016-08-05
Remove flier from boxplot on stripboxplot

arahuja:
This removes the outlier marker from `stripboxplot` since we plot each point anyways


tavinathanson:
LGTM

PR #104, opened 2016-08-05
Return namedtuples in plot functions

tavinathanson:
Main purpose: return more information for Mann Whitney, so I can easily compare the benefit vs. no-benefit distributions.

Also: use a customizable "alternative" for two-sided vs. one-sided (won't matter in the default cause, since `fishers_exact` defaults to `two-sided`).

@arahuja

### DIFF
--- a/cohorts/plot.py
+++ b/cohorts/plot.py
@@ -30,7 +30,7 @@ def stripboxplot(x, y, data, ax=None, **kwargs):
         x=x,
         y=y,
         data=data,
-        ax=ax, 
+        ax=ax,
         fliersize=0
     )
 
@@ -44,7 +44,7 @@ def stripboxplot(x, y, data, ax=None, **kwargs):
         **kwargs
     )
 
-def sided_str_from_alternative(alternative):
+def sided_str_from_alternative(alternative, condition):
     if alternative is None:
         raise ValueError("Must pick an alternative")
     if alternative == "two-sided":
@@ -65,7 +65,7 @@ def fishers_exact_plot(data, condition1, condition2, ax=None, alternative="two-s
         Dataframe to retrieve information from
 
     condition1: str
-        First binary column compare
+        First binary column to compare (and used for test sidedness)
 
     condition2: str
         Second binary column to compare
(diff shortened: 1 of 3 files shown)


cohorts/plot.py
```diff
@@ -43,7 +44,18 @@ def stripboxplot(x, y, data, ax=None, **kwargs):
...
 
-def fishers_exact_plot(data, condition1, condition2, ax=None):
+def sided_str_from_alternative(alternative):
+    if alternative is None:
+        raise ValueError("Must pick an alternative")
+    if alternative == "two-sided":
+        return alternative
+    # alternative hypothesis: condition is 'less' or 'greater' than no-condition
+    op_str = ">" if alternative == "greater" else "<"
+    return "one-sided: %s %s not %s" % (condition, op_str, condition)
```
arahuja, line 54: Where is the `condition` coming form?

tavinathanson, line 54: Yikes, that's confusing; especially since it was working as far as I could tell.

tavinathanson, line 54: Ah, it was working because I was only using one sided. 👊

arahuja, line 54: Did you hit this line ever? (Use a one-sided test?)

arahuja:
Test failure by the way, but seem unrelated.

Lint-ing would have caught it though:

```
************* Module cohorts.plot
E: 54,40: Undefined variable 'condition' (undefined-variable)
E: 54,59: Undefined variable 'condition' (undefined-variable)
```

tavinathanson:
@arahuja added in https://github.com/hammerlab/cohorts/pull/104/commits/0c6e9b060641b099ea7548ef7483c03b4de2145e:
- Pass `alternative` through in `load.py` for both Fisher and MW.
- Add `condition` to `sided_str_from_alternative`; for Fisher's, used `condition1`; I'm not sure if I'm thinking about this correctly. Relevant code: https://github.com/scipy/scipy/blob/v0.14.0/scipy/stats/stats.py#L2571
- Fixed broken test.

tavinathanson:
@arahuja for now I am just `ValueError`ing for `one-sided` Fisher's.

tavinathanson:
@arahuja good to merge?

arahuja:
👍

PR #111, opened 2016-08-10
Allow labeling of response/benefit labels

arahuja:
Converts the x-axis to look like
image

### DIFF
--- a/cohorts/load.py
+++ b/cohorts/load.py
@@ -1142,7 +1142,7 @@ class Cohort(Collection):
 
         condition_value = None
         if boolean_value_map:
-            assert (True in boolean_value_map) and (False in boolean_value_map), \
+            assert set(boolean_value_map.keys()) == set([True, False]), \
                 "Improper mapping of boolean column provided"
             df[boolean_col] = df[boolean_col].map(lambda v: boolean_value_map[v])
             condition_value = boolean_value_map[True]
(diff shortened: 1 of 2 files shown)


cohorts/load.py
```diff
@@ -1219,7 +1253,6 @@ def summarize_provenance_per_cache(self):
         frame for this cohort.
         """
         provenance_summary = {}
-        df = self.as_dataframe()
```
arahuja, line 1222: I was removing this as they showed up as unused, but now seeing that we are relying on a side-effect of these functions to fill in the value?

arahuja:
These PR originally dropped some used call to `as_dataframe()` but now seeing that we are relying on a side-effect of these functions to fill in the value?

cohorts/load.py
```diff
@@ -1114,7 +1135,18 @@ def plot_boolean(self, on, boolean_col, col=None, ax=None,
...
         df = filter_not_null(df, plot_col)
-        df[boolean_col] = df[boolean_col].astype(bool)
+
+        if boolean_label:
+            df[boolean_label] = df[boolean_col]
+            boolean_col = boolean_label
+
+        condition_value = None
+        if boolean_value_map:
+            assert (True in boolean_value_map) and (False in boolean_value_map), \
```
tavinathanson: Since you also want these to be the only keys, maybe `assert set(boolean_value_map.keys()) == set([True, False])`?

tavinathanson:
That seems confusing re `as_dataframe`. I'll file an issue. Otherwise LGTM!

arahuja:
Thanks, updated, the assert, will merge on Travis green-light

PR #112, opened 2016-08-10
Add p-value indicator

tavinathanson:
In this PR:
- Add random cohort generation to make it easy to test out plots without using private data.
- Add a significance indicator that shows \* if <= 0.05, otherwise "ns", for both Fisher's and MW.

image

image

image

@arahuja

### DIFF
--- a/cohorts/plot.py
+++ b/cohorts/plot.py
@@ -66,7 +66,7 @@ def add_significance_indicator(plot, col_a=0, col_b=1, significant=False):
     indicator = "*" if significant else "ns"
     plot.text((col_a + col_b) * 0.5, line_top, indicator, ha="center", va="bottom", color=color)
 
-def stripboxplot(x, y, data, ax=None, significant=False, **kwargs):
+def stripboxplot(x, y, data, ax=None, significant=None, **kwargs):
     """
     Overlay a stripplot on top of a boxplot.
     """
@@ -89,8 +89,10 @@ def stripboxplot(x, y, data, ax=None, significant=False, **kwargs):
         **kwargs
     )
 
-    hide_negative_y_ticks(plot)
-    add_significance_indicator(plot=plot, significant=significant)
+    if data[y].min() >= 0:
+        hide_negative_y_ticks(plot)
+    if significant is not None:
+        add_significance_indicator(plot=plot, significant=significant)
 
     return plot
 
--- a/cohorts/plot.py
+++ b/cohorts/plot.py
@@ -23,18 +23,31 @@ from sklearn.metrics import roc_curve
 from .model import bootstrap_auc
 
 def vertical_percent(plot, percent=0.1):
+    """
+    Using the size of the y axis, return a fraction of that size.
+    """
     plot_bottom, plot_top = plot.get_ylim()
     return percent * (plot_top - plot_bottom)
 
-def set_min_y_from_data(y_series, plot):
+def as_numeric(text):
+    try:
+        return float(text)
+    except:
+        return None
+
+def hide_negative_y_ticks(plot):
+    y_ticks = plot.get_yticks()
+    plot.set_yticks([tick for tick in y_ticks if as_numeric(tick) is not None and as_numeric(tick) >= 0])
+
+def only_percentage_ticks(plot):
     """
-    If data is >= 0, don't show negative axis values.
+    Only show ticks from 0.0 to 1.0.
     """
-    min_y_val = y_series.min()
-    if min_y_val >= 0:
-        # Add a bit of padding so that we don't cut the plot off in an ugly way
-        extra_space = -1 * vertical_percent(plot, 0.05)
-        plot.set_ylim(bottom=extra_space)
+    y_ticks = plot.get_yticks()
+    less_1_ticks = [tick for tick in y_ticks if as_numeric(tick) is not None and as_numeric(tick) <= 1]
+    if 1.0 not in less_1_ticks:
+        less_1_ticks.append(1.0)
+    plot.set_yticks(less_1_ticks)
 
 def add_significance_indicator(plot, col_a=0, col_b=1, significant=False):
     """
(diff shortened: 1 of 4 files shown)


cohorts/plot.py
```diff
@@ -22,7 +22,38 @@
 from sklearn.metrics import roc_curve
 from .model import bootstrap_auc
 
-def stripboxplot(x, y, data, ax=None, **kwargs):
+def vertical_percent(plot, percent=0.1):
+    plot_bottom, plot_top = plot.get_ylim()
+    return percent * (plot_top - plot_bottom)
+
+def set_min_y_from_data(y, data, plot):
+    """
```
arahuja, line 43: Does this need to take data and col name as an arg rather than the series itself?

cohorts/plot.py
```diff
@@ -22,7 +22,38 @@
...
+    If data is >= 0, don't show negative axis values.
+    """
+    min_y_val = data[y].min()
+    if min_y_val >= 0:
+        # Add a bit of padding so that we don't cut the plot off in an ugly way
+        extra_space = -1 * vertical_percent(plot, 0.05)
+        plot.set_ylim(bottom=extra_space)
```
arahuja: What I've done sometimes is turn off the bottom and top most tick label to have the extra space but not display negatives.

```
    yticks = plot.yaxis.get_major_ticks()
    yticks[0].set_visible(False)
    yticks[-1].set_visible(False)
```

Setting the max might be nice too, since sometimes we want things between 0 - 100%

arahuja:
Nice, like the random cohort. Only addition might making the significance level configurable, but not a must have.

cohorts/plot.py
tavinathanson: Interesting; maybe that's a more stable approach? Not sure the best way to automatically set the max; hard to know if we're talking about 0 to 100 or not?

tavinathanson: What would `yticks[-1]` do if the tick is `-0.5`?

arahuja: Actually looking at this again, not 100% positive I know what's happening. Mind explaining in person?

cohorts/plot.py
tavinathanson, line 43: Good point, fixed.

tavinathanson:
@arahuja I'm now using `set_yticks` and `get_yticks` to filter negative ticks, and the same strategy for non-0 to 1 ticks in the case of Fisher's exact. I imagine we can generalize the latter in a later PR, as well as let 0.05 be customizable.

image

image

tavinathanson:
@arahuja added a fix for the Fisher's exact ordering as well as custom styling in `styling.py`.

tavinathanson:
image

tavinathanson:
Also just simplified `__str__`s for `MannWhitneyResults` and Fisher's to not pollute notebooks.

tavinathanson:
@arahuja sorry to keep adding to this PR, but also just made it easier (https://github.com/hammerlab/cohorts/pull/112/commits/0d70a60c71d25d912ed6fd2dacc948f92787c344) to deal with situations where I want to split on something other than `benefit` that isn't presently in the `DataFrame`: I can include it as another function in `on`, and then specify which function is the `plot_col` vs. the `boolean_col`:

Example:

```
cohort.plot_boolean(on={"Liver Met": lambda row: row["Liver Mets"] == 1, "Split": split_patients},
                    plot_col="Liver Met",
                    boolean_col="Split")
```

cohorts/plot.py
```diff
@@ -45,6 +89,11 @@ def stripboxplot(x, y, data, ax=None, **kwargs):
         **kwargs
     )
 
+    hide_negative_y_ticks(plot)
```
arahuja: Are we always hiding negative ticks?

cohorts/plot.py
```diff
@@ -82,11 +137,16 @@ def fishers_exact_plot(data, condition1, condition2, ax=None, alternative="two-s
...
-        data=data
+        data=data,
+        **kwargs
     )
+
     count_table = pd.crosstab(data[condition1], data[condition2])
     print(count_table)
     oddsratio, pvalue = fisher_exact(count_table, alternative=alternative)
+    hide_negative_y_ticks(plot)
+    only_percentage_ticks(plot)
```
arahuja, line 150: Are these two always used in tandem? Or does `only_percentage_ticks` cover `hide_negative_y_ticks`? If not, should it?

arahuja:
Cool, the latest addition was something I would've found useful recently and may let me re-work how I did #111 to allow more customization. Will leave that off for now.

Left some general comments but makes sense if we want to get this in and come back to clean up some minor aspects.

arahuja:
Ah another thing I just realized, I use `stripboxplot` on its own frequently, with no stat. test, might be useful (in the future) to not plot the line, or `None` as an option for `significance`

cohorts/plot.py
tavinathanson, line 150: Great catch! Got lost in the fixes. Just re-added.

tavinathanson, line 150: Oops, that was intended for the above comment.

cohorts/plot.py
tavinathanson: Great catch! Got lost in the fixes. Just re-added.

cohorts/plot.py
tavinathanson, line 150: `only_percentage_ticks` is only used for Fisher's exact right now. Does that answer your question?

tavinathanson:
I think all comments are addressed and will plan to merge on 🍏  unless I hear anything!

cohorts/plot.py
arahuja, line 150: My question is:
- do you always need to call `hide_negative_y_ticks` before `only_percentage_ticks`?

If yes:
- should it just be called from `only_percentage_ticks`?
  If no:
- does `only_percentage_ticks` not remove negative ones on it's own?

tavinathanson, line 150: Oh. Right. Looks like that's confusingly named; perhaps I was intending them to be bundled but never did that, so `only_percentage_ticks` just handles <= 1. I just moved the Fisher's exact call to `hide_negative_y_ticks` into `only_percentage_ticks`.

PR #116, opened 2016-08-11
WIP: Prettify fn names

tavinathanson:
Rather than printing `missense_snv_col` in plots, this prints `Missense SNV Count / MB` or `Missense SNV Count`, depending on the params.

It's kind of the opposite of @jburos's renaming of columns, and can probably use some work to make that less confusing.

@arahuja

### DIFF
--- a/cohorts/utils.py
+++ b/cohorts/utils.py
@@ -15,6 +15,73 @@
 import re
 import warnings
 
+def first_not_none_param(params, default):
+    """
+    Given a list of `params`, use the first param in the list that is
+    not None. If all are None, fall back to `default`.
+    """
+    for param in params:
+        if param is not None:
+            return param
+    return default
+
+def filter_not_null(df, col):
+    original_len = len(df)
+    df = df[df[col].notnull()]
+    updated_len = len(df)
+    if updated_len < original_len:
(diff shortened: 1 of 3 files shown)


cohorts/utils.py
```diff
@@ -15,6 +15,73 @@
 import re
 import warnings
 
+def first_not_none_param(params, default):
```
tavinathanson, line 18: This is all copy/pasted; no changes.

tavinathanson:
@arahuja things are a little wonky so I'd actually hold off on reviewing.

tavinathanson:
This got a little complicated for the following reasons:
- Once I rename a column, something like `cohort.plot_boolean(on=[missense_snv_count, split_3_mos], plot_col="missense_snv_count", boolean_col="split_3_mos")` won't work because `"split_3_mos"` is renamed.
- It's non-trivial to figure out whether a function is normalizing per MB. I can hackishly inspect the arguments (`fn_argnames = fn.__code__.co_varnames; if "normalized_per_mb" in fn_argnames`), which works for e.g. `snv_count`, but that doesn't work for `expressed_neoantigen_count`, for which `normalized_per_mb` isn't actually an argument. I'll probably need to run the function and set state if we're normalizing.

Going to table this for now.

PR #117, opened 2016-08-12
Fix Fisher's exact plot for non-boolean columns

tavinathanson:
The switch from `benefit` to `Response`, combined with my `with_condition1_fraction` addition, didn't play well together; needed to use a `condition1_mask` rather than the same of the column.

@arahuja


arahuja:
ah, ok, I see the issue here, looks good, I'd say the alternative is to use the `count_table` to compute the means, but either works

tavinathanson:
@arahuja good call, sounds simpler, though I'll leave it as is for now.

PR #118, opened 2016-08-12
Always use a boolean mask

tavinathanson:
@arahuja this was the issue I was running into with:

```
condition_mask:
0     False
1     False
2      True
```

```
~condition_mask:
0     -1
1     -1
2     -2
```

Only happened for my `crpr` function that returned either `True`, `False` or `np.nan`. Pandas didn't set the dtype for that result to be `bool`, even though the `np.nan`s got filtered out before plotting, resulting in wonkiness.

I still don't understand why `~condition_mask` is even allowed for an `object` type?

### DIFF
--- a/cohorts/plot.py
+++ b/cohorts/plot.py
@@ -106,6 +106,17 @@ def sided_str_from_alternative(alternative, condition):
     op_str = ">" if alternative == "greater" else "<"
     return "one-sided: %s %s not %s" % (condition, op_str, condition)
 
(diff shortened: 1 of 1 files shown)


arahuja:
Is there a reason the function was returning `np.nan`? Do you want to force that to False or drop the row?

arahuja:
Oh ok I see, that the function can return any of the three, and then it is filtered, but the `dtype` is not set.

LGTM

PR #123, opened 2016-08-17
Fix survival plots for non-threshold conditions

arahuja:
Two changes here:
- labeling of the lines was broken when the condition passed was a binary series and not a comparison
  - added some configuration options while I was in there, allow the xlabel to be configurable (in case it needs to be months instead of days) and added a config option for `ci_show` out of curiosity.

### DIFF
--- a/cohorts/load.py
+++ b/cohorts/load.py
@@ -1182,7 +1182,7 @@ class Cohort(Collection):
                 ax=ax)
         return results
 
-    def plot_survival(self, on, col=None, how="os", ax=None,
+    def plot_survival(self, on, col=None, how="os", survival_units="Days", ax=None, ci_show=False,
                       threshold=None, **kwargs):
         """Plot a Kaplan Meier survival curve by splitting the cohort into two groups
         Parameters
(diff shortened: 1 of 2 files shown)


tavinathanson:
LGTM, thanks for fixing that! For consistency, `label_no_condition` should probably be `not` the other one vs. empty?

I'd be okay with making whether or not we want two labels vs. one (one implied) configurable, but I think the two types (threshold and non-threshold) should be consistent with each other?

arahuja:
>  For consistency, label_no_condition should probably be not the other one vs. empty?

I do agree with this, but the not is not always obvious? For example, if the condition was 'contains x, y or z' we don't know what the alternatives are. We could do what we had previously which was literally append 'NOT _' in front, but that seemed weird, not clean for actual paper figures.

I agree on having them both do the same thing. The right thing to do here to just allow the labels to be set be the user in this case? But I didn't want to design around that for now and just wanted a fix?

If that sounds good, we can open an issue and come back to a real solution?

tavinathanson:
We talked about using a non-English operator like ø or ¬ or ~.

arahuja:
Added ¬

PR #124, opened 2016-08-17
Add survival legend label option

e5c:
This enables the legend to have a custom label e.g. "APOBEC Signature" instead of using `condition_col` (in my case, "signatures_normalized_value")


tavinathanson:
@e5c why not just pass in `col="APOBEC Signature"` into `plot_survival`?

PR #126, opened 2016-08-22
Add default survival colors

tavinathanson:
Open to other color suggestions!

image

@arahuja

### DIFF
--- a/cohorts/survival.py
+++ b/cohorts/survival.py
@@ -29,8 +29,8 @@ def plot_kmf(df,
              xlabel=None,
              ylabel=None,
              ax=None,
-             with_condition_color=colors.hex2color("#5B3636"),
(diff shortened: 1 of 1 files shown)


cohorts/survival.py
```diff
@@ -50,6 +53,8 @@ def plot_kmf(df,
                                   at its median
         title: Title for the plot, default None
         ax: an existing matplotlib ax, optional, default None
+        with_condition_color: tuple, color for the with-condition curve
```
arahuja: Clarify that is an RGB tuple or maybe take the hexcode directly?

arahuja:
👍  No strong preferences on colors, I like that it is configurable.

cohorts/survival.py
tavinathanson: Good call, going with the latter approach.

PR #127, opened 2016-08-23
Allow configurable labels to survival plots

arahuja:


cohorts/survival.py
```diff
@@ -67,14 +69,17 @@ def plot_kmf(df,
...
-        label_with_condition = "%s > %s" % (condition_col, label_suffix)
+        default_label_with_condition = "%s > %s" % (condition_col, label_suffix)
     else:
         condition = df[condition_col]
-        label_with_condition = "= {}".format(condition_col)
-        label_no_condition = "¬ {}".format(condition_col)
+        default_label_with_condition = "= {}".format(condition_col)
+        default_label_no_condition = "¬ {}".format(condition_col)
+
+    with_condition_label = with_condition_label or default_label_with_condition
```
tavinathanson, line 81: Weird, didn't realize this was valid syntax.

tavinathanson:
LGTM!

cohorts/survival.py
arahuja: Just looked it up actually, was tired of `x if x is not None else ...` syntax

PR #129, opened 2016-08-24
Fix fishers plot ranges

arahuja:
Current Fisher's Exact plot can end up as below if the values are closer to 0:

image

This fixes them so it looks like:
image

Or like this if the values are closer to 1
image

Also, changing the way we turn of negative values for yticks allows matplotlib to change the labels appropriately without squeezing labels in

i.e. this can happen now

image

where this will happen now
image

### DIFF
--- a/cohorts/plot.py
+++ b/cohorts/plot.py
@@ -35,20 +35,23 @@ def as_numeric(text):
     except:
         return None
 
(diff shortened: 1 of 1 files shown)


tavinathanson:
Nice! I thought I had an issue with using `set_visible`, but since I can't remember what it was at the moment, LGTM!

PR #132, opened 2016-09-08
Support joining on DataFrameLoaders with shared columns

tavinathanson:
For example:

`join_with=["loader_a", "loader_b"]`

Where `loader_a` has columns `patient_id` `A` `B` `C` and `loader_b` has columns `patient_id` `C` `D`...

...will result in columns like:

```
`patient_id`
`A`
`B`
`loader_a_C`
`loader_b_C`
`D`
```

@arahuja

### DIFF
--- a/cohorts/load.py
+++ b/cohorts/load.py
@@ -40,6 +40,7 @@ from topiary.sequence_helpers import contains_mutant_residues
 from isovar.protein_sequence import variants_to_protein_sequences_dataframe
 from pysam import AlignmentFile
 from scipy.stats import pearsonr
+from collections import defaultdict
 
 from .utils import strip_column_names as _strip_column_names
 from .survival import plot_kmf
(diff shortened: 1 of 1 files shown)


arahuja:
Wouldn't passing `suffixes=('', '_' + df_loader.name)` to `merge` accomplish the same? Or is that different somehow?

arahuja:
Wonder why the multiple (and disagreeing) coverage messages on the same commit

tavinathanson:
@arahuja very strange indeed re the coverage.

tavinathanson:
@arahuja re `suffixes`, that's what I started with, but it gets problematic with multiple `DataFrameLoader`s. Since I don't know a better way to join > 2 `DataFrame`s together other than sequential joins, the first join would result in non-prefixed or suffixed columns, while the second would get the suffix.

In the above example, that'd result in `loader_a_C` and `C` rather than `loader_a_C` and `loader_b_C`.

tavinathanson:
@arahuja I just switched the prefixes to suffixes for consistency with `pandas`.

arahuja:
Ah, thanks for the explanation.

LGTM - on a minor note, I don't think `rename` errors on missing columns, so you could find all the missing columns at once and then bulk rename the columns, though not clear that offers an advantage.

tavinathanson:
@arahuja yeah, I was thinking about doing that but I think it's probably fine as is?

PR #134, opened 2016-09-09
Add barplot option to plot_joint and change plot_joint name to plot_correlation

tavinathanson:
Keeping `plot_joint` around, but deprecated, for now.

@arahuja

### DIFF
--- a/cohorts/load.py
+++ b/cohorts/load.py
@@ -1266,7 +1266,7 @@ class Cohort(Collection):
         warnings.warn("plot_joint is deprecated; use plot_correlation.", Warning)
         return self.plot_correlation(*args, **kwargs)
 
-    def plot_correlation(self, on, on_two=None, x_col=None, plot_type="jointplot", stat_func=pearsonr, show_stat_func=True, jointplot_kwargs={}, **kwargs):
+    def plot_correlation(self, on, on_two=None, x_col=None, plot_type="jointplot", stat_func=pearsonr, show_stat_func=True, plot_kwargs={}, **kwargs):
         """Plot the correlation between two variables.
 
         Parameters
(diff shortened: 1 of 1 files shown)


cohorts/load.py
```diff
@@ -1262,7 +1262,11 @@ def plot_survival(self,
         )
         return results
 
-    def plot_joint(self, on, on_two=None, x_col=None, plot_type="jointplot", stat_func=pearsonr, show_stat_func=True, jointplot_kwargs={}, **kwargs):
+    def plot_joint(self, *args, **kwargs):
+        warnings.warn("plot_joint is deprecated; use plot_correlation.", Warning)
```
tavinathanson, line 1266: Wanted this to be `DeprecationWarning`, but that doesn't show by default in notebooks and I'm not clear what the best way is to change that setting. Open to suggestions!

cohorts/load.py
```diff
@@ -1282,7 +1286,7 @@ def plot_joint(self, on, on_two=None, x_col=None, plot_type="jointplot", stat_fu
         jointplot_kwargs : dict, optional
```
arahuja: Should this be specific to `jointplot` or just have `plot_kwargs` passed to all the plots? i.e. passing `ax` might be useful

tavinathanson: That seems like a good idea. Will make an issue for it unless you think it should happen in this PR?

tavinathanson: On second thought, easy enough to do here.

tavinathanson:
@arahuja I generalized `jointplot_kwargs` to `plot_kwargs`.

arahuja:
👍  LGTM

Issue #37, opened 2016-05-18
Break up load.py


tavinathanson:
It's a bit monstrous at the moment.

e5c:
(Happy to take this on as part of the public-facing example usage effort)

tavinathanson:
Sure, let us know when you're actually moving all the code around so that we can try to avoid crazy merge conflicts.

tavinathanson:
I think a good first step would be moving `Patient`-related functions into the `Patient` object and out of the `Cohort` object. Then, (ideally in a separate PR) `Patient` can be moved to `patient.py`. Caching logic should probably end up in a `cache.py`, though I think separate PRs for each step would be best.

arahuja:
@tavinathanson more to do here after #138? Maybe open an issue with more specifics now that `load.py` is gone?

tavinathanson:
I'm okay to close this for now.

PR #138, opened 2016-09-16
Split up load.py

tavinathanson:
Only moves/renames in this PR, to start shrinking `load.py`. Plenty more needs to happen; this is just a start on https://github.com/hammerlab/cohorts/issues/37.

@arahuja


cohorts/patient.py
```diff
@@ -0,0 +1,108 @@
...
+                 benefit=None,
+                 snv_vcf_paths=[],
+                 indel_vcf_paths=[],
+                 normal_sample=None,
+                 tumor_sample=None,
+                 hla_alleles=None,
+                 additional_data=None,
+                 cohort=None):
+        require_id_str(id)
+        self.id = id
```
arahuja, line 69: Not necessary here, but saw a nice library to simplify things like this https://attrs.readthedocs.io/en/stable/examples.html

arahuja:
👍

PR #139, opened 2016-09-19
Fix travis run

tavinathanson:
In this PR:
- Don't use the latest `pyensembl` since we're not using the latest `varcode`.
- Use a different `matplotlib` backend.

@arahuja


arahuja:
👍  assuming all is well with Travis

PR #150, opened 2016-09-28
Make API more consistent

tavinathanson:
In this PR:
- The default for `as_dataframe` is to always return just the `DataFrame`, unless `return_cols` is `True`. I use a `DataFrameHolder` `namedtuple` that's generally invisible to the user, in order to pass `return_cols` around and return the right thing.
- `col` is removed from `plot_benefit`, `plot_survival`, etc.
- The `on` API is standardized as `str`/function/list of `str`s or functions/dict of `str`s to `str`s or functions; `plot_benefit`, `plot_survival`, `plot_correlation`, etc. obey this.
- `plot_joint` is now removed in favor of `plot_correlation`. `plot_correlation` uses the `on` API to specify two variables rather than `on_two`.
- Added `coxph`, `bootstrap_auc` and `mean_bootstrap` AUC to the `cohorts` object, and they use the same `on` API.

Examples of the new API in action:

```
df = cohort.as_dataframe(on=missense_snv_count)
cols, df = cohort.as_dataframe(on={"aa": "2-factor score", "bb": missense_snv_count}, return_cols=True)
df = cohort.as_dataframe({"hello": "2-factor score", "goodbye": missense_snv_count})
cohort.plot_correlation(on={"aa": "2-factor score", "bb": missense_snv_count})
cohort.plot_correlation(on=["2-factor score", missense_snv_count])
cohort.mean_bootstrap_auc(on="Age")
cohort.mean_bootstrap_auc(on=missense_snv_count)
`cohort.coxph(on=[missense_snv_count, "Age"])`
```

Hopefully we can get this in before we acquire any new users. I figure this'll be `0.4.0`.

@arahuja cc @jburos

### DIFF
--- a/cohorts/cohort.py
+++ b/cohorts/cohort.py
@@ -1126,6 +1126,8 @@ class Cohort(Collection):
         if plot_type not in ["boxplot", "barplot", "jointplot"]:
             raise ValueError("Invalid plot_type %s" % plot_type)
         plot_cols, df = self.as_dataframe(on, return_cols=True, **kwargs)
+        if len(plot_cols) != 2:
+            raise ValueError("Must be comparing two columns, but there are %d columns" % len(plot_cols))
         for plot_col in plot_cols:
             df = filter_not_null(df, plot_col)
         if x_col is None:
--- a/cohorts/cohort.py
+++ b/cohorts/cohort.py
@@ -47,6 +47,7 @@ from .utils import DataFrameHolder, first_not_none_param, filter_not_null, Inval
 from .provenance import compare_provenance
 from .survival import plot_kmf
 from .plot import mann_whitney_plot, fishers_exact_plot, roc_curve_plot, stripboxplot, CorrelationResults
+from .model import cohort_coxph, cohort_bootstrap_auc, cohort_mean_bootstrap_auc
 from .collection import Collection
 from .varcode_utils import (filter_variants, filter_effects,
                             filter_neoantigens, filter_polyphen)
@@ -1104,19 +1105,13 @@ class Cohort(Collection):
         )
         return results
 
-    def plot_joint(self, *args, **kwargs):
-        warnings.warn("plot_joint is deprecated; use plot_correlation.", Warning)
-        return self.plot_correlation(*args, **kwargs)
-
-    def plot_correlation(self, on, on_two=None, x_col=None, plot_type="jointplot", stat_func=pearsonr, show_stat_func=True, plot_kwargs={}, **kwargs):
+    def plot_correlation(self, on, x_col=None, plot_type="jointplot", stat_func=pearsonr, show_stat_func=True, plot_kwargs={}, **kwargs):
         """Plot the correlation between two variables.
 
         Parameters
         ----------
-        on : function or list or dict of functions
+        on : list or dict of functions or strings
             See `cohort.load.as_dataframe`
-        on_two : function, optional
-            Can specify the second function here rather than creating a list.
         x_col : str, optional
             If `on` is a dict, this guarantees we have the expected ordering.
         plot_type : str, optional
@@ -1130,9 +1125,7 @@ class Cohort(Collection):
         """
         if plot_type not in ["boxplot", "barplot", "jointplot"]:
             raise ValueError("Invalid plot_type %s" % plot_type)
-        if on_two is not None:
-            on = [on, on_two]
-        plot_cols, df = self.as_dataframe(on, **kwargs)
+        plot_cols, df = self.as_dataframe(on, return_cols=True, **kwargs)
         for plot_col in plot_cols:
             df = filter_not_null(df, plot_col)
         if x_col is None:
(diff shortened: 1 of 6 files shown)


cohorts/cohort.py
```diff
@@ -300,22 +295,23 @@ def apply_func(on, col, df):
             (as `self`) along if the function accepts a `cohort` argument.
             """
             if col is None:
-                # Use the function name, or "column" for lambdas, if no
-                # name is provided for the newly created column.
-                col = on.__name__ if not is_lambda(on) else "column"
+                col = func_name(on)
```
arahuja: Do we test this ever? `func_name` seems like is defined after this point and inside of this function?

cohorts/cohort.py
```diff
@@ -326,20 +322,27 @@ def is_lambda(func):
...
             cols = []
             for key, value in on.iteritems():
-                col, df = apply_func(on=value, col=key, df=df)
+                if type(value) == str:
+                    df[key] = df[value]
+                    col = key
+                elif type(value) == FunctionType:
+                    col, df = apply_func(on=value, col=key, df=df)
                 cols.append(col)
         if type(on) == list:
```
arahuja, line 331: `elif`? Shouldn't be possible for both to be true?

cohorts/cohort.py
```diff
@@ -300,22 +295,23 @@ def apply_func(on, col, df):
...
             if "cohort" not in on_argnames:
                 func = lambda row: on(row=row, **kwargs)
             else:
                 func = lambda row: on(row=row, cohort=self, **kwargs)
             df[col] = df.apply(func, axis=1)
-            return (col, df)
+            return DataFrameHolder(col, df)
+
+        def func_name(func, num=0):
+            return func.__name__ if not is_lambda(func) else "column_%d" % i
```
arahuja: Should this be using `num`?

cohorts/utils.py
```diff
@@ -16,6 +16,22 @@
...
+    """Holds a DataFrame along with associated columns of interest."""
+    def return_self(self, return_cols):
+        """Either return the DataFrame or the DataFrameHolder."""
+        if return_cols:
+            return self
+        else:
+            return self.df
+
+    @staticmethod
+    def return_obj(cols, df, return_cols=False):
```
arahuja, line 31: I don't think I see why this is a `namedtuple` with this create method instead of just a class with `init`?

arahuja (changes requested):
Overall LGTM, there is one place where I think there is an error and then some minor comments

cohorts/utils.py
tavinathanson: I still use it as a `tuple` in a bunch of places; made for a less intensive refactor. Does that seem reasonable?

tavinathanson: It can be simply `init`ed with `DataFrameHolder(cols, df)`; I use that in a few places; but this `return_obj` method is to further allow _either_ the df or the holder object to be returned depending on the value of `return_cols`. Just a shortcut since that needs to happen a fair amount.

cohorts/cohort.py
tavinathanson: It's defined outside this function, so should work? That said, don't think anyone calls/needs this, so removing this line.

cohorts/cohort.py
tavinathanson: Yikes, good catch. Definitely want more tests soon.

cohorts/cohort.py
tavinathanson: Also want to error if neither, adding that.

tavinathanson: That said, don't use `elif` is wrong here even if unnecessary?

tavinathanson:
@arahuja I tried to address your comments; additionally I added `plot_col_from_cols` and used it in a few functions to allow an `on` = `dict` or `list` to work when it contains a single column without additionally specifying the `plot_col` (which is needed when `on` contains multiple functions).

Example of what wasn't possible before, and is now:

```
cohort.plot_survival(["Age"])
cohort.plot_survival({"aa": "Age"})
cohort.plot_benefit(["Age"])
cohort.plot_benefit({"aa": "Age"})
```

cohorts/cohort.py
arahuja: 👍  Think I was just reading it wrong on github

cohorts/cohort.py
arahuja: > don't use elif is wrong

Yea not wrong, just wanted to make sure I understood it right and only one was possible.

cohorts/cohort.py
```diff
@@ -1014,15 +1034,8 @@ def plot_boolean(self,
...
-        if type(cols) == str:
-            if plot_col is not None:
-                raise ValueError("plot_col is specified when it isn't ndeeded.")
-            plot_col = cols
-        elif type(cols) == list:
-            if plot_col is None:
-                raise ValueError("plot_col must be specified when multiple `on`s are present.")
-
+        cols, df = self.as_dataframe(on, return_cols=True, **kwargs)
+        plot_col = self.plot_col_from_cols(cols=cols, plot_col=plot_col)
```
arahuja, line 1040: No need to take plot_col as an argument then? It is always overridden?

cohorts/cohort.py
```diff
@@ -924,12 +927,33 @@ def cohort_columns(self):
         column_types = [cohort_dataframe[col].dtype for col in cohort_dataframe.columns]
         return dict(zip(list(cohort_dataframe.columns), column_types))
 
-    def plot_roc_curve(self, on, bootstrap_samples=100, col=None, ax=None, **kwargs):
+    def plot_col_from_cols(self, cols, only_allow_one=True, plot_col=None):
+        if type(cols) == str:
+            if plot_col is not None:
+                raise ValueError("plot_col is specified when it isn't ndeeded because there is only one col.")
```
arahuja: nit: spelling `ndeeded`

cohorts/cohort.py
arahuja: Sorry, that's not true - but when would you need to specify`plot_col`?

arahuja (approved):
LGTM overall - I think I'm still not positive when we hit all the cases provided, so as we refine this perhaps some comments with an example of when we would end up there?

cohorts/cohort.py
tavinathanson: Fixing

cohorts/cohort.py
tavinathanson: Good question: we might want to use `on` to grab not just e.g. `snv_count` in `cohort.plot_benefit(on=snv_count)`, but actually to grab the `benefit_col` itself; sometimes we want to split patients by a column that isn't already in the `DataFrame`. I'll mention this here?

Issue #96, opened 2016-07-26
Simplify Cohort default


tavinathanson:
Seems very easy to accidentally forget the following lines at the beginning of functions:

```
filter_fn = first_not_none_param([filter_fn, cohort.filter_fn], no_filter)
normalized_per_mb = first_not_none_param([normalized_per_mb, cohort.normalized_per_mb], False)
```

...and thereby not use the `Cohort` default.

tavinathanson:
Indeed, @jburos ran into this very issue: https://github.com/hammerlab/cohorts/issues/120#issue-171240320

PR #158, opened 2016-11-01
Simplify functions with decorators

tavinathanson:
This is my first attempt to address https://github.com/hammerlab/cohorts/issues/96: namely, to ensure that `filter_fn` and `normalized_per_mb` obey the `Cohort`-level defaults.

Functions now handle this via a decorator called `use_defaults`. Count functions use another decorator, `count_function`, which takes advantage of `use_defaults` but also adds a bunch of counting logic that can be factored out of the other functions.

Also in this PR:
* Got rid of `functions.py` in the `test` directory. They overrode the actual `functions.py` with `normalized_per_mb = False`, but I realized this was easy enough to do in e.g. `test_basic.py` by setting `Cohort.normalized_per_mb = False`.
* Add more relevant tests.

I'm open to alternative ideas; not wed to this. @jburos what do you think?


test/test_count.py
```diff
@@ -203,69 +203,3 @@ def test_multiple_effects():
             rmtree(vcf_dir)
         if cohort is not None:
             cohort.clear_caches()
-
```
tavinathanson, line 206: Moved, not deleted.

jburos (approved):
Really nice updates -- LGTM. I feel like this would still benefit from a short how-to, but it will be much simpler to write. I will work on it.

tavinathanson:
@jburos cool, thanks! And agreed that we still need to explain a bit :)

PR #159, opened 2016-11-04
Add regplot as a plot_correlation option

tavinathanson:
`regplot` does less than `jointplot` does, and also only uses one `ax` rather than multiple, so it can be placed inside a `subplot`. Long story short: I'd like to use `regplot` today :)

@jburos


jburos (approved):
LGTM

PR #165, opened 2016-12-16
Add more granular variant specification and other minor changes

tavinathanson:
Added a few things to support some of my melanoma and lung work.

In this PR:

* Biggest change is removing `snv_vcf_paths` and `indel_vcf_paths` and replacing with a dictionary from `snv` to a list of VCF paths or MAF paths or `VariantCollection`s.
* Allow `filter`ing of a `Cohort`.
* Allow users to grab *all* effects, not just top priority ones.
* Add a `.genome` property.

Since removing `snv_vcf_paths` will break things, I'd be bumping the release up to `0.5.0`. Also need to do more manual/automated testing to confirm no issues.


jburos:
First, thanks for working on this & helping to tackle these issues!

Few high-level comments:

1. Not sure I like the dict approach to indel, snv or any. Why not include all VCF/MAFs irrespective of type & handle the filtering by type in a function? (e.g. have `variant_count`, `snv_count`, `indel_count`, etc)?
2. Also, not sure I like breaking backward compatibility for this reason. I think we can add an option for `vcf_paths` (to accept a list) and populate this with values from `snv_vcf_paths` & `indel_vcf_paths` if the user passes these in?
3. I do think the behavior of `variant-type` default behavior is a little tricky to reconcile without modifying the default filter fn. How wedded to this are you?

I would like to test this out if that's OK with you & will then give more concrete feedback.

jburos:
Just noting here our offline discussion: 

1. remove dict of vcf files; replace with list. 
2. remove variant_type param; filter types using summary functions.
3. OK to break backwards compatibility.

Otherwise LGTM --

tavinathanson:
@jburos responded to all the above, and added an associated test as well as an error if a cache with the old format is present. I don't want to bother you over the weekend, so I'll probably go ahead and merge today or tomorrow.

jburos:
LGTM - thanks Tavi! Really like the out-of-date cache check.

Issue #167, opened 2016-12-19
Support "merged" strelka VCF type


jburos:
Previously cohorts accepted a `snv_vcf_list` and an `indel_vcf_list`, but our current epidisco runs output a merged vcf. 

There are two things that would help support this better:

1. Change vcf input to a list taking any variant collection file type
2. modify snv_count & related functions to filter variants by type, irrespective of source file identified

jburos:
Part of this is supported via PR #165

jburos:
Also related to PR #169

Issue #166, opened 2016-12-17
Using only Strelka tier 1 results in divide by zero sometimes


tavinathanson:
@jburos:

> any idea why I'd see a variant with no ref/alt reads at all? Getting a divide by zero error in vaf calc :confused:
```(Pdb) variant
Variant(contig='13', start=111589766, ref='C', alt='T', reference_name='GRCh37')
(Pdb) sample_info
OrderedDict([('AU', [0, 0]), ('CU', [0, 0]), ('DP', 10), ('FDP', 0), ('GU', [10, 48]), ('SDP', 0), ('SUBDP', 0), ('TU', [0, 1])])
(Pdb) ref_depth
0
(Pdb) int(sample_info[variant.alt+"U"][0])
0
```

From @tavinathanson:

> maybe it is what i thought @ihodes was saying, namely a bug in cohorts since we only grab tier 1, which can sometimes result in 0; so maybe we should filter those out (still weird to me that we didn't hit this before is the thing)

From @jburos:

> was thinking to return depth of 0 & set the vaf to 0 for now; users can/should filter them through qc filters (we could mod the no_filter filter to remove them)

From @tavinathanson:
> i actually don't think `no_filter` should ever do anything, arun was a strong proponent of eventually getting rid of it which i support; added it in for a few reasons though so it's extraction would need to be thoughtful

From @jburos:
>right now the load_variant functions don't do any filtering, which is (i think) expected behavior. so for now i'll leave this & not filter on loading variant_stats .. best practice should be for users to set their own qc filters

From @tavinathanson:
>maybe return depth of 0 and VAF of `None`? 0 for VAF seems wrongish; agree that filters can take it from there

From @jburos:
> sounds good - was thinking `None` as well but didn't know how many things this might break

jburos:
Also related to PR #169

PR #169, opened 2016-12-19
Fix variant stats; cached filtered variants; add exonic counts; add progress bar

### LINKED ISSUE #170: Cache sometimes fails if user interrupts writing process
This probably only ever happens to me, but the cache is sometimes invalidated when I "interrupt" a process which is writing to the cache, and then try again. 

The output (after resuming an interrupted process) would look something like : 

```
/home/jacquelineburos/projects/cohorts/cohorts/cohort.py in load_from_cache(self, cache_name, patient_id, file_name)
    427         else:
    428             with open(cache_file, "rb") as f:
--> 429                 return pickle.load(f)
    430 
    431     def save_to_cache(self, obj, cache_name, patient_id, file_name):

EOFError: ('Ran out of input', 'occurred at index 23')
```

There are two ways to address this : 

1. catch the interruption & don't write to cache unless it finishes.
2. catch the error reading from cache & handle it gracefully. 

The second seems more foolproof to me so this is where I'll start. But good to keep both in mind.

jburos:
This is an updated version of another PR - #168

Notes from that PR:

1. Addresses issues #166 - variant stats lead to divide by zero error when total read count is 0 for a variant
2. Also addresses #167 - filtering variant stats by type, irrespective of merged file type
3. caches filtered variants as well as unfiltered variant collections
4. pass **kwargs to filter_fn, if provided. This enables more flexible filtering 

Also now includes PR #171:

1. Clean up some issues in filtered-variants cache, namely:
    - inspect closure of filter_fn for other filter_fns, use these to name the cached file
    - ^ above method only works for python3; skip filtered cache on py2
    - added some tests to identify issues in the above
2. Made things more robust to errors reading from cache - e.g. fix #170  
    - skip reading from cache if there is any error
3. added a number of new summary functions to `cohorts.functions`, including `exonic_snv_count`, `expressed_exonic_snv_count`, etc.
3. functions named 'snv' now explicitly filter for `.is_snv` type
4. pass **kwargs through summary functions to filter functions; e.g. for functions like `variant_qc_filter`, so that the following works for `as_dataframe` & `plot_benefit`, etc.:

    ```
    df = cohort.as_dataframe(
            on=[exonic_snv_count],
            filter_fn=cohorts.variant_filters.variant_qc_filter,
            min_tumor_depth=7,
            min_normal_depth=7,
            min_tumor_vaf=0.1,
            max_normal_vaf=0.03,
            min_tumor_alt_depth=0)
    ```
5. allow **kwargs use with multiple functions, e.g. so the above can be multiplexed:
    ```
    df = cohort.as_dataframe(
            on=[exonic_snv_count, missense_snv_count, expressed_missense_snv_count],
            filter_fn=cohorts.variant_filters.variant_qc_filter,
            min_tumor_depth=7,
            min_normal_depth=7,
            min_tumor_vaf=0.1,
            max_normal_vaf=0.03,
            min_tumor_alt_depth=0)
    ```
    This will print a "warning" statement that the kwargs are passed to all functions, in lieu of previous behavior which resulted in an error.
7. Added debug logging statements throughout load-variants & application of filter functions to test behavior.
8. Added progress bar printing progress of function-application, to better know if process is running or stalled. 
    - modified logging to not use stdout, so that progress bars print correctly.

    Output looks like:
    screen shot 2016-12-21 at 2 08 07 pm

### DIFF
--- a/cohorts/cohort.py
+++ b/cohorts/cohort.py
@@ -49,7 +49,7 @@ from scipy.stats import pearsonr
 from collections import defaultdict
 
 from .dataframe_loader import DataFrameLoader
-from .utils import DataFrameHolder, first_not_none_param, filter_not_null, InvalidDataError, strip_column_names as _strip_column_names
+from .utils import DataFrameHolder, first_not_none_param, filter_not_null, InvalidDataError, strip_column_names as _strip_column_names, get_logger
 from .provenance import compare_provenance
 from .survival import plot_kmf
 from .plot import mann_whitney_plot, fishers_exact_plot, roc_curve_plot, stripboxplot, CorrelationResults
@@ -61,7 +61,7 @@ from .variant_filters import no_filter
 from .styling import set_styling
 from . import variant_filters
 
-logger = logging.getLogger(__name__)
+logger = get_logger(__name__, level=logging.INFO)
 
 class Cohort(Collection):
     """
@@ -375,6 +375,7 @@ class Cohort(Collection):
         Instead of joining a DataFrameJoiner with the Cohort in `as_dataframe`, sometimes
         we may want to just directly load a particular DataFrame.
         """
+        logger.debug('loading dataframe: {}'.format(df_loader_name))
         # Get the DataFrameLoader object corresponding to this name.
         df_loaders = [df_loader for df_loader in self.df_loaders if df_loader.name == df_loader_name]
 
@@ -402,11 +403,14 @@ class Cohort(Collection):
         if not self.cache_results:
             return None
         
+        logger.debug('loading patient {} data from {} cache: {}'.format(patient_id, cache_name, file_name))
+
         cache_dir = path.join(self.cache_dir, cache_name)
         patient_cache_dir = path.join(cache_dir, str(patient_id))
         cache_file = path.join(patient_cache_dir, file_name)
 
         if not path.exists(cache_file):
+            logger.debug('... file does not exist. Checking for older format.')
             # We removed variant_type from the cache name. Eventually remove this notification.
             if (path.exists(path.join(patient_cache_dir, "snv-" + file_name)) or
                 path.exists(path.join(patient_cache_dir, "indel-" + file_name))):
@@ -414,6 +418,7 @@ class Cohort(Collection):
             return None
 
         if self.check_provenance:
+            logger.debug('... Checking cache provenance')
             num_discrepant = compare_provenance(
                 this_provenance = self.generate_provenance(),
                 other_provenance = self.load_provenance(patient_cache_dir),
@@ -422,8 +427,10 @@ class Cohort(Collection):
                 )
         try:
             if path.splitext(cache_file)[1] == ".csv":
+                logger.debug('... Loading cache as csv file')
                 return pd.read_csv(cache_file, dtype={"patient_id": object})
             else:
+                logger.debug('... Loading cache as pickled file')
                 with open(cache_file, "rb") as f:
                     return pickle.load(f)
         except:
@@ -432,6 +439,8 @@ class Cohort(Collection):
     def save_to_cache(self, obj, cache_name, patient_id, file_name):
         if not self.cache_results:
             return
+        
+        logger.debug('saving patient {} data to {} cache: {}'.format(patient_id, cache_name, file_name))
 
         cache_dir = path.join(self.cache_dir, cache_name)
         patient_cache_dir = path.join(cache_dir, str(patient_id))
@@ -478,6 +487,11 @@ class Cohort(Collection):
             Dictionary of patient_id to VariantCollection
         """
         filter_fn = first_not_none_param([filter_fn, self.filter_fn], no_filter)
+        if filter_fn is None:
+            filter_fn_name = 'None'
+        else:
+            filter_fn_name = filter_fn.__name__
+        logger.debug('loading variants with filter_fn: {}'.format(filter_fn_name))
         patient_variants = {}
 
         for patient in self.iter_patients(patients):
@@ -490,6 +504,7 @@ class Cohort(Collection):
     def _hash_filter_fn(self, filter_fn, **kwargs):
         if filter_fn is None:
             return 'none'
+        logger.debug('Computing hash for filter_fn: {} with kwargs {}'.format(filter_fn_name, str(dict(**kwargs))))
         # function source code
         fn_source = str(dill.source.getsource(filter_fn))
         hashed_fn_source = pickle.dumps(fn_source)
@@ -499,7 +514,7 @@ class Cohort(Collection):
         if not kw_dict:
             kw_hash = ['default']
         else:
-            [kw_hash.append('{}-{}'.format(key, h)) for (key, h) in kw_dict.items()]
+            [kw_hash.append('{}-{}'.format(key, h)) for (key, h) in sorted(kw_dict.items())]
             
         # closure vars
         closure = ''
@@ -515,13 +530,19 @@ class Cohort(Collection):
         
     
     def _load_single_patient_variants(self, patient, filter_fn, use_cache=True, **kwargs):
+        if filter_fn is None:
+            filter_fn_name = 'None'
+        else:
+            filter_fn_name = filter_fn.__name__
+        logger.debug('loading variants for patient {} with filter_fn {}'.format(patient.id, filter_fn_name))
         use_filtered_cache = use_cache
         if sys.version_info < (3, 3):
-            logger.debug('Disabling filtered cache due to python version')
+            logger.debug('... disabling filtered cache due to python version')
             use_filtered_cache = False
             
         ## get cache name, if possible
         if use_filtered_cache:
+            logger.debug('... identifying filtered-cache file name')
             try:
                 ## try to load filtered variants from cache
                 filtered_cache_file_name = "%s-variants.filter-%s.pkl" % (self.merge_type,
@@ -531,6 +552,7 @@ class Cohort(Collection):
         
         ## read from cache, if possible
         if use_filtered_cache:
+            logger.debug('... trying to load filtered variants from cache: {}'.format(filtered_cache_file_name))
             try:
                 cached = self.load_from_cache(self.cache_names["variant"], patient.id, filtered_cache_file_name)
                 if cached is not None:
@@ -540,25 +562,29 @@ class Cohort(Collection):
                 pass
         
         ## get merged variants
+        logger.debug('... getting merged variants for: {}'.format(patient.id))
         merged_variants = self._load_single_patient_merged_variants(patient, use_cache=use_cache)
 
         # Note that this is the number of variant collections and not the number of
         # variants. 0 variants will lead to 0 neoantigens, for example, but 0 variant
         # collections will lead to NaN variants and neoantigens.
         if merged_variants is None:
-            print("Variants did not exist for patient %s" % patient.id)
+            logger.info("Variants did not exist for patient %s" % patient.id)
             return None
         
+        logger.debug('... applying filters to variants for: {}'.format(patient.id))
         filtered_variants = filter_variants(variant_collection=merged_variants,
                                             patient=patient,
                                             filter_fn=filter_fn,
                                             **kwargs)
         if use_filtered_cache:
+            logger.debug('... saving filtered variants to cache: {}'.format(filtered_cache_file_name))
             self.save_to_cache(filtered_variants, self.cache_names["variant"], patient.id, filtered_cache_file_name)
         return filtered_variants
 
     
     def _load_single_patient_merged_variants(self, patient, use_cache=True):
+        logger.debug('loading merged variants for patient {}'.format(patient.id))
         failed_io = False
         try:
             # get merged-variants from cache
@@ -608,6 +634,7 @@ class Cohort(Collection):
     
     
     def _merge_variant_collections(self, variant_collections, merge_type):
+        logger.debug('Merging variants using merge type: {}'.format(merge_type))
         assert merge_type in ["union", "intersection"], "Unknown merge type: %s" % merge_type
         head = variant_collections[0]
         if merge_type == "union":
@@ -723,6 +750,12 @@ class Cohort(Collection):
              Dictionary of patient_id to varcode.EffectCollection
         """
         filter_fn = first_not_none_param([filter_fn, self.filter_fn], no_filter)
+        if filter_fn is None:
+            logger.warn('Loading patient effects with filter_fn: None')
+            filter_fn_name = 'None'
+        else:
+            filter_fn_name = filter_fn.__name__
+        logger.debug('loading effects with filter_fn {}'.format(filter_fn_name))
         patient_effects = {}
         for patient in self.iter_patients(patients):
             effects = self._load_single_patient_effects(
@@ -733,6 +766,11 @@ class Cohort(Collection):
 
     def _load_single_patient_effects(self, patient, only_nonsynonymous, all_effects, filter_fn):
         cached_file_name = "%s-effects.pkl" % self.merge_type
+        if filter_fn is None:
+            filter_fn_name = 'None'
+        else:
+            filter_fn_name = filter_fn.__name__
+        logger.debug('loading effects for patient {} with filter_fn {}'.format(patient.id, filter_fn_name))
 
         # Don't filter here, as these variants are used to generate the
         # effects cache; and cached items are never filtered.
--- a/cohorts/cohort.py
+++ b/cohorts/cohort.py
@@ -506,6 +506,7 @@ class Cohort(Collection):
     def _hash_filter_fn(self, filter_fn, **kwargs):
         if filter_fn is None:
             return 'filter-none'
+        filter_fn_name = filter_fn.__name__
         logger.debug('Computing hash for filter_fn: {} with kwargs {}'.format(filter_fn_name, str(dict(**kwargs))))
         # function source code
         fn_source = str(dill.source.getsource(filter_fn))
--- a/cohorts/cohort.py
+++ b/cohorts/cohort.py
@@ -556,7 +556,7 @@ class Cohort(Collection):
             logger.info("... disabling filtered cache due to python version")
             use_filtered_cache = False
             
-        ## get cache name, if possible
+        ## confirm that we can get cache-name (else don't use filtered cache)
         if use_filtered_cache:
             logger.debug("... identifying filtered-cache file name")
             try:
@@ -564,18 +564,18 @@ class Cohort(Collection):
                 filtered_cache_file_name = "%s-variants.%s.pkl" % (self.merge_type,
                                                                    self._hash_filter_fn(filter_fn, **kwargs))
             except:
+                logger.warning("... error identifying filtered-cache file name for patient {}: {}".format(
+                        patient.id, filter_fn_name))
                 use_filtered_cache = False
-        
-        ## read from cache, if possible
-        if use_filtered_cache:
-            logger.debug("... trying to load filtered variants from cache: {}".format(filtered_cache_file_name))
-            try:
-                cached = self.load_from_cache(self.cache_names["variant"], patient.id, filtered_cache_file_name)
-                if cached is not None:
-                    return cached
-            except:
-                logger.warn("Error loading variants from cache for patient: {}".format(patient.id))
-                pass
+            finally:
+                logger.debug("... trying to load filtered variants from cache: {}".format(filtered_cache_file_name))
+                try:
+                    cached = self.load_from_cache(self.cache_names["variant"], patient.id, filtered_cache_file_name)
+                    if cached is not None:
+                        return cached
+                except:
+                    logger.warning("Error loading variants from cache for patient: {}".format(patient.id))
+                    pass
         
         ## get merged variants
         logger.debug("... getting merged variants for: {}".format(patient.id))
--- a/cohorts/cohort.py
+++ b/cohorts/cohort.py
@@ -567,7 +567,7 @@ class Cohort(Collection):
                 logger.warning("... error identifying filtered-cache file name for patient {}: {}".format(
                         patient.id, filter_fn_name))
                 use_filtered_cache = False
-            finally:
+            else:
                 logger.debug("... trying to load filtered variants from cache: {}".format(filtered_cache_file_name))
                 try:
                     cached = self.load_from_cache(self.cache_names["variant"], patient.id, filtered_cache_file_name)
--- a/cohorts/cohort.py
+++ b/cohorts/cohort.py
@@ -505,7 +505,7 @@ class Cohort(Collection):
     
     def _hash_filter_fn(self, filter_fn, **kwargs):
         if filter_fn is None:
-            return 'none'
+            return 'filter-none'
         logger.debug('Computing hash for filter_fn: {} with kwargs {}'.format(filter_fn_name, str(dict(**kwargs))))
         # function source code
         fn_source = str(dill.source.getsource(filter_fn))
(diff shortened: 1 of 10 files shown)


hammer:
@jburos would you mind given a more detailed name to this PR? The name shows up on task boards and "Develop" doesn't give me a good idea of what's going on.

jburos:
Sure - apologies for strange naming; will copy description & other text from the related PR

cohorts/cohort.py
```diff
@@ -56,6 +61,8 @@
 from .styling import set_styling
 from . import variant_filters
 
+logger = logging.getLogger(__name__)
```
tavinathanson: Should use `get_logger`, will be fixed in other PR

cohorts/functions.py
```diff
@@ -67,13 +68,65 @@ def get_patient_to_mb(cohort):
...
+    # This only loads one effect per variant.
+    patient_id = row["patient_id"]
+    return cohort.load_effects(
+        only_nonsynonymous=False,
+        patients=[cohort.patient_from_id(patient_id)],
+        filter_fn=exonic_filter_fn,
+        **kwargs)
+
 @count_function
 def snv_count(row, cohort, filter_fn, normalized_per_mb, **kwargs):
```
tavinathanson, line 94: We talked in person about possibly simplifying this one, but probably best for another PR that simplifies all the functions a bit more.

cohorts/functions.py
```diff
@@ -67,13 +68,65 @@ def get_patient_to_mb(cohort):
...
+    patient_id = row["patient_id"]
+    def indel_filter_fn(filterable_variant, **kwargs):
+        assert filter_fn is not None, "filter_fn should never be None, but it is."
+        return (filterable_variant.variant.is_indel and
+                filter_fn(filterable_variant=filterable_variant, **kwargs))
+    return cohort.load_variants(
+        patients=[cohort.patient_from_id(patient_id)],
+        filter_fn=indel_filter_fn,
+        **kwargs)
+
```
tavinathanson, line 116: Remove two extra newlines?

cohorts/functions.py
```diff
@@ -67,13 +68,65 @@ def get_patient_to_mb(cohort):
...
+
+
+@count_function
+def effect_count(row, cohort, filter_fn, normalized_per_mb, **kwargs):
```
tavinathanson, line 118: What's the intent, here? By default, I think (but may be wrong) that this should be the same as `variant_count` since by default we only return one highest priority effect per variant. However, if you pass in `all_effects=True` you'll get something different?

cohorts/functions.py
```diff
@@ -67,13 +68,65 @@ def get_patient_to_mb(cohort):
...
+    ''' Alternate count of variants using effects instead of variants '''
```
tavinathanson, line 119: Style: docstrings should always use " vs. '

cohorts/functions.py
```diff
@@ -100,6 +154,66 @@ def missense_filter_fn(filterable_effect):
         **kwargs)
 
 @count_function
+def exonic_snv_count(row, cohort, filter_fn, normalized_per_mb, **kwargs):
+    def exonic_filter_fn(filterable_effect, **kwargs):
```
tavinathanson: Not that it matters, but for consistency maybe call this `exonic_snv_filter_fn`?

cohorts/functions.py
```diff
@@ -100,6 +154,66 @@ def missense_filter_fn(filterable_effect):
...
+    # This only loads one effect per variant.
+    patient_id = row["patient_id"]
+    return cohort.load_effects(
+        only_nonsynonymous=False,
+        patients=[cohort.patient_from_id(patient_id)],
+        filter_fn=exonic_filter_fn,
+        **kwargs)
+
+@count_function
+def exonic_indel_count(row, cohort, filter_fn, normalized_per_mb, **kwargs):
```
tavinathanson: I think this should be `deletion` vs. `indel`?

cohorts/functions.py
tavinathanson: Especially as the number of functions is ballooning for everything possible combination, this seems worth solving.

tavinathanson: Added https://github.com/hammerlab/cohorts/issues/173

cohorts/functions.py
```diff
@@ -100,6 +154,66 @@ def missense_filter_fn(filterable_effect):
...
+    # This only loads one effect per variant.
+    patient_id = row["patient_id"]
+    return cohort.load_effects(
+        only_nonsynonymous=True,
+        patients=[cohort.patient_from_id(patient_id)],
+        filter_fn=exonic_indel_filter_fn,
+        **kwargs)
+
+@count_function
+def exonic_insert_count(row, cohort, filter_fn, normalized_per_mb, **kwargs):
```
tavinathanson: Call this `insertion`?

cohorts/utils.py
```diff
@@ -179,11 +179,11 @@ def strip_column_names(cols, keep_paren_contents=True):
...
 
-def get_logger(name):
+def get_logger(name, level=logging.INFO):
     logger = logging.getLogger(name)
     if logger.handlers:
         logger.handlers = []
-    stdout_handler = logging.StreamHandler(sys.stdout)
-    logger.addHandler(stdout_handler)
-    logger.setLevel(logging.INFO)
+    #stdout_handler = logging.StreamHandler(sys.stdout)
```
tavinathanson: Remove comments?

cohorts/variant_stats.py
```diff
@@ -27,7 +27,9 @@ def strelka_somatic_variant_stats(variant, variant_metadata):
     sample_info = variant_metadata["sample_info"]
     # Ensure there are exactly two samples in the VCF, a tumor and normal
     assert len(sample_info) == 2, "More than two samples found in the somatic VCF"
+
```
tavinathanson: Unnecessary 3 newlines?

cohorts/variant_stats.py
```diff
@@ -45,12 +47,23 @@ def _strelka_variant_stats(variant, sample_info):
...
-
-    # Retrieve the Tier 1 counts from Strelka
-    ref_depth = int(sample_info[variant.ref+"U"][0])
-    alt_depth = int(sample_info[variant.alt+"U"][0])
-    depth = alt_depth + ref_depth
-    vaf = float(alt_depth) / depth
+    
+    if variant.is_deletion or variant.is_insertion:
+        # ref: https://sites.google.com/site/strelkasomaticvariantcaller/home/somatic-variant-output
+        #depth = int(sample_info['DP']) # read depth for tier 1
```
tavinathanson: Remove this line?

cohorts/variant_stats.py
```diff
@@ -45,12 +47,23 @@ def _strelka_variant_stats(variant, sample_info):
...
+        # not sure why this depth is different from the sum of ref + alt
+        ref_depth = int(sample_info['TAR'][0]) # number of reads supporting ref allele (non-deletion)
+        alt_depth = int(sample_info['TIR'][0]) # number of reads supporting alt allele (deletion)
+        depth = ref_depth + alt_depth
```
tavinathanson, line 49: You mentioned that DP doesn't always add up to ref + alt depth; can you make an issue describing that so we can look into it more?

cohorts/variant_stats.py
```diff
@@ -45,12 +47,23 @@ def _strelka_variant_stats(variant, sample_info):
...
+    else:
+        # Retrieve the Tier 1 counts from Strelka
+        ref_depth = int(sample_info[variant.ref+"U"][0])
+        alt_depth = int(sample_info[variant.alt+"U"][0])
+        depth = alt_depth + ref_depth
+    if depth > 0:
+        vaf = float(alt_depth) / depth
+    else:
+        vaf = None
```
tavinathanson, line 60: I think this makes sense, as we've discussed, though remind me: where is this `None` handled downstream?

tavinathanson: Ah, right, I guess it's up to the person writing the QC filter to filter these out. Can you add a comment clarifying that?

test/test_cohort_defaults.py
```diff
@@ -32,7 +32,44 @@
 
 FILE_FORMAT_1 = "patient_format1_%s.vcf"
 
-def test_default_filter_fn():
+def test_default_filter_fn_with_variant_count():
```
tavinathanson: Does this mostly repeat the `snv_count` one? Can they just be the same code called twice?

cohorts/cohort.py
```diff
@@ -470,23 +489,117 @@ def load_variants(self, patients=None, filter_fn=None):
...
+        fn_source = str(dill.source.getsource(filter_fn))
+        pickled_fn_source = pickle.dumps(fn_source)
+        # kwarg values
+        kw_dict = dict(**kwargs)
+        kw_hash = list()
+        if not kw_dict:
+            kw_hash = ['default']
+        else:
+            [kw_hash.append('{}-{}'.format(key, h)) for (key, h) in sorted(kw_dict.items())]
+        # closure vars
```
tavinathanson: Where's the test that breaks if you don't have this?

cohorts/cohort.py
```diff
@@ -470,23 +489,117 @@ def load_variants(self, patients=None, filter_fn=None):
...
+            if inspect.isfunction(val):
+                closure = '{}-{}'.format(self._hash_filter_fn(val).__name__, self._hash_filter_fn(val))
+        # construct final hashed_fn
+        hashed_fn = '.'.join(['-'.join([filter_fn.__name__,
+                                        int(hashlib.sha1(pickled_fn_source).hexdigest(), 16) % (10 ** 11)]),
+                              '.'.join(kw_hash),
+                              closure]
+                            )
+        return hashed_fn
+
```
tavinathanson: Excess newline?

cohorts/cohort.py
```diff
@@ -319,7 +327,8 @@ def apply_func(on, col, df):
                 func = lambda row: on(row=row, **kwargs)
             else:
                 func = lambda row: on(row=row, cohort=self, **kwargs)
-            df[col] = df.apply(func, axis=1)
+            tqdm.pandas(desc=col)
+            df[col] = df.progress_apply(func, axis=1)
```
tavinathanson: Comment here that this relies on `tqdm` above?

cohorts/cohort.py
```diff
@@ -369,6 +377,7 @@ def load_dataframe(self, df_loader_name):
         Instead of joining a DataFrameJoiner with the Cohort in `as_dataframe`, sometimes
         we may want to just directly load a particular DataFrame.
         """
+        logger.debug('loading dataframe: {}'.format(df_loader_name))
```
tavinathanson: Since this file uses " vs. ' usually, use " here and in some of you other changes?

cohorts/cohort.py
```diff
@@ -395,35 +404,45 @@ def save_provenance(self, patient_cache_dir, provenance):
...
             return None
+        
+        logger.debug('loading patient {} data from {} cache: {}'.format(patient_id, cache_name, file_name))
 
         cache_dir = path.join(self.cache_dir, cache_name)
         patient_cache_dir = path.join(cache_dir, str(patient_id))
         cache_file = path.join(patient_cache_dir, file_name)
 
         if not path.exists(cache_file):
+            logger.debug('... file does not exist. Checking for older format.')
```
tavinathanson: Can you say `cache file`?

cohorts/cohort.py
```diff
@@ -395,35 +404,45 @@ def save_provenance(self, patient_cache_dir, provenance):
...
-                return pickle.load(f)
+        try:
+            if path.splitext(cache_file)[1] == ".csv":
+                logger.debug('... Loading cache as csv file')
+                return pd.read_csv(cache_file, dtype={"patient_id": object})
+            else:
+                logger.debug('... Loading cache as pickled file')
+                with open(cache_file, "rb") as f:
+                    return pickle.load(f)
+        except:
```
tavinathanson: Maybe just except `IOError` or something to allow other errors to bubble up?

cohorts/cohort.py
```diff
@@ -470,23 +489,117 @@ def load_variants(self, patients=None, filter_fn=None):
...
+        logger.debug('loading variants with filter_fn: {}'.format(filter_fn_name))
         patient_variants = {}
 
         for patient in self.iter_patients(patients):
-            variants = self._load_single_patient_variants(patient, filter_fn)
+            variants = self._load_single_patient_variants(patient, filter_fn, **kwargs)
             if variants is not None:
                 patient_variants[patient.id] = variants
         return patient_variants
+
```
tavinathanson, line 510: Remove newline?

cohorts/cohort.py
```diff
@@ -470,23 +489,117 @@ def load_variants(self, patients=None, filter_fn=None):
             Dictionary of patient_id to VariantCollection
         """
         filter_fn = first_not_none_param([filter_fn, self.filter_fn], no_filter)
+        if filter_fn is None:
```
tavinathanson: One line? `filter_fn_name = filter_fn.__name__ if filter_fn is not None else "None"`?

cohorts/cohort.py
```diff
@@ -470,23 +489,117 @@ def load_variants(self, patients=None, filter_fn=None):
...
+    
+    
+    def _hash_filter_fn(self, filter_fn, **kwargs):
+        if filter_fn is None:
+            return 'filter-none'
+        filter_fn_name = filter_fn.__name__
+        logger.debug('Computing hash for filter_fn: {} with kwargs {}'.format(filter_fn_name, str(dict(**kwargs))))
+        # function source code
+        fn_source = str(dill.source.getsource(filter_fn))
+        pickled_fn_source = pickle.dumps(fn_source)
```
tavinathanson: Mention why pickling here?

cohorts/cohort.py
tavinathanson: Can you also mention the missense_snv_count example and why this is necessary in a comment?

cohorts/cohort.py
```diff
@@ -638,16 +755,27 @@ def load_effects(self, patients=None, only_nonsynonymous=False,
              Dictionary of patient_id to varcode.EffectCollection
         """
         filter_fn = first_not_none_param([filter_fn, self.filter_fn], no_filter)
+        if filter_fn is None:
```
tavinathanson: In fact, this code is repeated enough that I think it should be extracted into a `get_filter_fn_name(default_name="None")` function?

cohorts/cohort.py
tavinathanson: Actually, this seems to happen 4+ times, so I'd extract into a function as I mentioned again below

cohorts/cohort.py
```diff
@@ -470,23 +489,117 @@ def load_variants(self, patients=None, filter_fn=None):
...
+        
+    
+    def _load_single_patient_variants(self, patient, filter_fn, use_cache=True, **kwargs):
+        if filter_fn is None:
+            filter_fn_name = 'None'
+        else:
+            filter_fn_name = filter_fn.__name__
+        logger.debug('loading variants for patient {} with filter_fn {}'.format(patient.id, filter_fn_name))
+        use_filtered_cache = use_cache
+        if sys.version_info < (3, 3):
```
tavinathanson, line 555: Made https://github.com/hammerlab/cohorts/issues/174

cohorts/cohort.py
```diff
@@ -470,23 +489,117 @@ def load_variants(self, patients=None, filter_fn=None):
...
+            logger.debug('... disabling filtered cache due to python version')
```
tavinathanson: This should probably be more than `DEBUG` message here?

cohorts/cohort.py
```diff
@@ -470,23 +489,117 @@ def load_variants(self, patients=None, filter_fn=None):
...
+            use_filtered_cache = False
+            
+        ## get cache name, if possible
+        if use_filtered_cache:
+            logger.debug('... identifying filtered-cache file name')
+            try:
+                ## try to load filtered variants from cache
+                filtered_cache_file_name = "%s-variants.%s.pkl" % (self.merge_type,
+                                                                          self._hash_filter_fn(filter_fn, **kwargs))
+            except:
```
tavinathanson, line 566: Why two `try`s? Why not just continue into the ## read form cache within the same `try`?

cohorts/cohort.py
```diff
@@ -500,29 +613,33 @@ def _load_single_patient_variants(self, patient, filter_fn):
                     variant_collections.append(patient_variants)
                 else:
                     raise ValueError("Don't know how to read %s" % patient_variants)
+            # merge variant-collections
+            if len(variant_collections) == 0:
+                failed_io = True
```
tavinathanson: Since this isn't actually failed IO, rename the variable to `no_variants` or something?

cohorts/cohort.py
```diff
@@ -470,23 +489,117 @@ def load_variants(self, patients=None, filter_fn=None):
...
+        
+        ## get merged variants
+        logger.debug('... getting merged variants for: {}'.format(patient.id))
+        merged_variants = self._load_single_patient_merged_variants(patient, use_cache=use_cache)
 
-    def _load_single_patient_variants(self, patient, filter_fn):
+        # Note that this is the number of variant collections and not the number of
+        # variants. 0 variants will lead to 0 neoantigens, for example, but 0 variant
+        # collections will lead to NaN variants and neoantigens.
+        if merged_variants is None:
```
tavinathanson, line 585: I'm unclear why this 5-line block needs to be repeated if it's present in `_load_single_patient_variants`?

cohorts/cohort.py
```diff
@@ -470,23 +489,117 @@ def load_variants(self, patients=None, filter_fn=None):
...
+        hashed_fn = '.'.join(['-'.join([filter_fn.__name__,
+                                        int(hashlib.sha1(pickled_fn_source).hexdigest(), 16) % (10 ** 11)]),
+                              '.'.join(kw_hash),
+                              closure]
+                            )
+        return hashed_fn
+            
+        
+    
+    def _load_single_patient_variants(self, patient, filter_fn, use_cache=True, **kwargs):
```
tavinathanson, line 543: Add comment/docstrings stating that this function deals with loading/saving to the filter_fn cache, while the merged variant function has the unfiltered variant cache?

cohorts/cohort.py
```diff
@@ -470,23 +489,117 @@ def load_variants(self, patients=None, filter_fn=None):
...
+        if merged_variants is None:
+            logger.info("Variants did not exist for patient %s" % patient.id)
+            return None
+        
+        logger.debug('... applying filters to variants for: {}'.format(patient.id))
+        filtered_variants = filter_variants(variant_collection=merged_variants,
+                                            patient=patient,
+                                            filter_fn=filter_fn,
+                                            **kwargs)
+        if use_filtered_cache:
```
tavinathanson, line 594: This is all a nice reminder that caching could be made more seamless https://github.com/hammerlab/cohorts/issues/175

tavinathanson (changes requested):
Looks good! Got some comments but thanks for all these additions!

test/test_cohort_defaults.py
jburos: no, because one (`missense_snv_count`) will break without the 'closure' cache check & the other does not. it helps to have both tests in there to see where the failure occurs.

tavinathanson: I don't mean remove the test, I mean don't duplicate the code?

jburos: I see - thanks. i'll reformat

cohorts/cohort.py
jburos: INFO?

cohorts/cohort.py
jburos: I tried to find most of these -- i have the opposite convention but will try to adjust

cohorts/cohort.py
jburos: This is part of the effort to make the caching "safer" -- see #170 

Short answer is, the two "try" statements are capturing different issues. 
* first: If the name of the cache can't be determined we shouldn't try to load or save from cache. Here we set `use_filtered_cache = False`.
* second: If there is an error reading from cache, we should re-compute filtered variants & (try to) overwrite cache file. Here we do not set `use_filtered_cache = False`. Instead we just keep going to compute filtered variants.

cohorts/cohort.py
jburos: I re-edited it to clarify that we want to preserve the value of `None` as being qualitatively different from `0`.

cohorts/cohort.py
jburos: Agreed. this was a holdover from earlier code

cohorts/functions.py
jburos: This is a sanity check, to test the assumption that variant_count & effect_count should be identical. Also a good way to make sure that filter_fns are applied as expected.

cohorts/cohort.py
tavinathanson: My attempted solution to endless " vs. ' debates is to at least try to preserve consistency within files; so if another file is mostly ' vs. " I'm fine with leaving that convention.

cohorts/functions.py
tavinathanson: Can you add a docstring explaining the expected use case to avoid confusion in the future between the two?

cohorts/cohort.py
tavinathanson: Maybe even `WARNING`?

cohorts/cohort.py
tavinathanson: Maybe at least put the second `try` inside the first `try` to avoid two successive `if use_filtered_cache:` statements, which I find confusing?

cohorts/cohort.py
tavinathanson: 👍  yeah, previously `failed_io` did indicate only failed IO, so then I think it made sense; but now it indicates something more broad

cohorts/variant_stats.py
jburos: If/when I find an example of this, I will include it.

cohorts/cohort.py
jburos: That's not equivalent logical structure - making it equivalent will end up with what (IMO) will end up being a more confusing set of nested Try & If statements.  You can think of the second `if` as a "if we are *still* using filtered cache".

tavinathanson:
Understood; what I'm getting at is that changing the meaning of
`use_filtered_cache` and then re-checking it again below seems
confusing/error-prone.

But if it's not obvious how to improve it, I'm fine with leaving it.

On Tue, Jan 17, 2017 at 3:44 PM, Jacki Buros Novik <notifications@github.com

cohorts/cohort.py
tavinathanson: Understood; what I'm getting at is that changing the meaning of
`use_filtered_cache` and then re-checking it again below seems
confusing/error-prone.

But if it's not obvious how to improve it, I'm fine with leaving it.

jburos:
I'll try to clarify the intention of `use_filtered_cache` in the comments.

jburos:
@tavinathanson made a bunch of requested changes - can you take another look when you have a chance? Not urgent ..

tavinathanson:
@jburos probably won't get to this until next week if that's okay!

test/test_cohort_defaults.py
```diff
@@ -32,9 +32,9 @@
 
 FILE_FORMAT_1 = "patient_format1_%s.vcf"
 
-def test_default_filter_fn():
+def test_default_filter_fn(summary_fn=variant_count):
```
tavinathanson, line 35: Does this test run at all for `variant_count` now that it has a parameter? Just a question of how `nose` works; don't recall.

cohorts/variant_stats.py
```diff
@@ -1,14 +1,12 @@
 from collections import namedtuple
-
+import numpy as np
```
tavinathanson: Is this used?

tavinathanson:
@jburos added two minor comments but feel free to merge once rebased, sorry for the delay!

test/test_cohort_defaults.py
jburos: I don't know - i believe you had asked me to make this a function. I will force a failure to test.

tavinathanson: I asked to not duplicate code, that's right. I'm just asking if you need another function called `test_default_filter_fn_variant_count` that calls this for it to run.

jburos: To test this I ran nosetests with the decorator `@nottest` in front of that function & without the decorator. 35 tests ran with the decorator, whereas without the decorator 36 tests ran. So I think it does run. LMK if you think this test should be done differently.

cohorts/variant_stats.py
jburos: This is leftover from the earlier version of the divide-by-zero issue, which did use np. I'll remove it.

jburos:
Merged with commit 26e935d

PR #171, opened 2016-12-21
Add exonic count functions

### LINKED ISSUE #170: Cache sometimes fails if user interrupts writing process

jburos:
Modifications:

1. Clean up some issues in filtered-variants cache, namely:
    - inspect closure of filter_fn for other filter_fns, use these to name the cached file
    - ^ above method only works for python3; skip filtered cache on py2
    - added some tests to identify issues in the above
2. Made things more robust to errors reading from cache - e.g. fix #170  
    - skip reading from cache if there is any error
3. added a number of new summary functions to `cohorts.functions`, including `exonic_snv_count`, `expressed_exonic_snv_count`, etc.
3. functions named 'snv' now explicitly filter for `.is_snv` type
4. pass **kwargs through summary functions to filter functions; e.g. for functions like `variant_qc_filter`, so that the following works for `as_dataframe` & `plot_benefit`, etc.:

    ```
    df = cohort.as_dataframe(
            on=[exonic_snv_count],
            filter_fn=cohorts.variant_filters.variant_qc_filter,
            min_tumor_depth=7,
            min_normal_depth=7,
            min_tumor_vaf=0.1,
            max_normal_vaf=0.03,
            min_tumor_alt_depth=0)
    ```
5. allow **kwargs use with multiple functions, e.g. so the above can be multiplexed:
    ```
    df = cohort.as_dataframe(
            on=[exonic_snv_count, missense_snv_count, expressed_missense_snv_count],
            filter_fn=cohorts.variant_filters.variant_qc_filter,
            min_tumor_depth=7,
            min_normal_depth=7,
            min_tumor_vaf=0.1,
            max_normal_vaf=0.03,
            min_tumor_alt_depth=0)
    ```
    This will print a "warning" statement that the kwargs are passed to all functions, in lieu of previous behavior which resulted in an error.
7. Added debug logging statements throughout load-variants & application of filter functions to test behavior.
8. Added progress bar printing progress of function-application, to better know if process is running or stalled. 
    - modified logging to not use stdout, so that progress bars print correctly.

    Output looks like:
    screen shot 2016-12-21 at 2 08 07 pm

### DIFF
--- a/cohorts/cohort.py
+++ b/cohorts/cohort.py
@@ -49,7 +49,7 @@ from scipy.stats import pearsonr
 from collections import defaultdict
 
 from .dataframe_loader import DataFrameLoader
-from .utils import DataFrameHolder, first_not_none_param, filter_not_null, InvalidDataError, strip_column_names as _strip_column_names
+from .utils import DataFrameHolder, first_not_none_param, filter_not_null, InvalidDataError, strip_column_names as _strip_column_names, get_logger
 from .provenance import compare_provenance
 from .survival import plot_kmf
 from .plot import mann_whitney_plot, fishers_exact_plot, roc_curve_plot, stripboxplot, CorrelationResults
@@ -61,7 +61,7 @@ from .variant_filters import no_filter
 from .styling import set_styling
 from . import variant_filters
 
-logger = logging.getLogger(__name__)
+logger = get_logger(__name__, level=logging.INFO)
 
 class Cohort(Collection):
     """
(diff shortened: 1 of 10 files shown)


jburos:
This PR is being merged into #169 -- closing.

Issue #176, opened 2017-01-16
Add isovar usage tests


tavinathanson:
Will require a test BAM.

PR #177, opened 2017-01-16
Upgrade isovar to the latest version in cohorts

tavinathanson:
In this PR:
* Upgrade isovar from 0.0.6 to the latest version
* Fix `expressed_neoantigen_count` to pass `normalized_per_mb` and `filter_fn` through; this shouldn't have impacted our analyses unless we ever passed in a `filter_fn` directly into `expressed_neoantigen_count` (versus using the cohort-level default).
* Fix `expressed_variant_set` memoizing such that it memoizes per-cohort (ran into an issue where I had two `Cohort`s and it was sharing the value.

This could use some tests; filed https://github.com/hammerlab/cohorts/issues/176

FYI @iskandr in case you see anything wonky

### DIFF
--- a/cohorts/cohort.py
+++ b/cohorts/cohort.py
@@ -38,7 +38,8 @@ from varcode import EffectCollection, VariantCollection
 from mhctools import NetMHCcons, EpitopeCollection
 from topiary import predict_epitopes_from_variants, epitopes_to_dataframe
 from topiary.sequence_helpers import contains_mutant_residues
-from isovar.protein_sequence import variants_to_protein_sequences_dataframe
+from isovar.allele_reads import reads_overlapping_variants
+from isovar.protein_sequences import reads_generator_to_protein_sequences_generator, protein_sequences_generator_to_dataframe
 from pysam import AlignmentFile
 from scipy.stats import pearsonr
 from collections import defaultdict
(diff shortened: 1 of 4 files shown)


cohorts/cohort.py
```diff
@@ -940,15 +937,18 @@ def load_single_patient_isovar(self, patient, variants, epitope_lengths):
...
             samfile=rna_bam_file,
+            min_mapping_quality=1)
+        protein_sequences_generator = reads_generator_to_protein_sequences_generator(
+            allele_reads_generator,
             protein_sequence_length=protein_sequence_length,
-            min_reads_supporting_rna_sequence=3, # Per Alex R.'s suggestion
-            min_transcript_prefix_length=MIN_TRANSCRIPT_PREFIX_LENGTH,
-            max_transcript_mismatches=MAX_REFERENCE_TRANSCRIPT_MISMATCHES,
+            # Per Alex R.'s suggestion; equivalent to min_reads_supporting_rna_sequence previously
+            min_variant_sequence_coverage=3,
```
iskandr, line 948: Revised opinion is that it's probably safe to use 2 here.

(Unsourced intuition)

jburos (approved):
LGTM - happy to see this update

tavinathanson:
@iskandr gotcha, good to know, leaving it as 3 for now to avoid confusion with previous analyses

PR #179, opened 2017-01-17
Grab VAFs from MAFs

tavinathanson:
Grabs depth/VAF information from MAFs, assuming a particular MAF format.

This is pending https://github.com/hammerlab/varcode/pull/209, and the `requirements.txt` might have to be adjusted accordingly.


jburos (approved):
LGTM Tavi - seems useful. thanks.

jburos:
We should probably add some tests for these functions, but could be left as a separate issue

tavinathanson:
@jburos agreed, rushing a bit to make some progress on Lung but will file an issue

Issue #182, opened 2017-02-17
deletion_count is counting all indels


tavinathanson:
`is_indel` is used in `deletion_count`; the function should be renamed to `indel_count`.

tavinathanson:
Fixed by #186

Issue #181, opened 2017-02-17
exonic_silent_snv_count is confusing


tavinathanson:
I would think this means only silent mutations; but the code looks like it includes both silent and non-silent.

tavinathanson:
Fixed by #186

PR #186, opened 2017-02-23
Refactor and fix functions

### LINKED ISSUE #184: Minor: old comment
This comment is no longer applicable in `cohort.py`:

```
# For multiple functions, don't allow kwargs since we won't know which functions
# they apply to
```

tavinathanson:
In this PR:

* Added `count_effects_function_builder` and `count_variants_function_builder` to get rid of a bunch of boilerplate.
* Tried to make our functions consistent. For example, if a function is counting nonsynonymous mutations, that should be clear in the title (where I think `missense` is clearly a subset). I renamed `exonic_insertion_count` and `exonic_deletion_count` to `nonsynonymous_insertion_count` and `nonsynonymous_deletion_count`, since exonic is implied when we're talking about nonsynonymous mutations, and those functions had `only_nonsynonymous=True`.
   * On a related note: does this mean that our exonic SNV counts in the bladder paper are actually nonsynonymous mutation counts? Maybe not, since https://github.com/hammerlab/bladder-analyses/blob/e6a48c04d4d49a1af10fe89e1b563731440ae841/analyses/utils/extra_functions.py#L4 doesn't appear to filter by nonsynonymous?
* Added more functions.
   * New `exonic_` functions that act differently: `only_nonsynonymous=False`, so that they're not just nonsynonymous mutations.
   * Renamed `exonic_silent_snv_count` to just `exonic_snv_count`, per the above (and https://github.com/hammerlab/cohorts/issues/181).
   * Added `indel_count`, `insertion_count`, `deletion_count`; including fixing the existing `deletion_count` (https://github.com/hammerlab/cohorts/issues/182).
* Tried to add unit tests to compare new vs. old, but it got unwieldy. Instead, I just made a notebook where I copied all the old functions and compared with the new counterparts: https://github.com/hammerlab/msk-bms-lung/blob/master/analyses/notebooks/Test%20New%20Count%20Functions.ipynb

Also fixes https://github.com/hammerlab/cohorts/issues/184

### DIFF
--- a/cohorts/functions.py
+++ b/cohorts/functions.py
@@ -68,150 +68,128 @@ def get_patient_to_mb(cohort):
     patient_to_mb = dict(cohort.as_dataframe(join_with="ensembl_coverage")[["patient_id", "MB"]].to_dict("split")["data"])
     return patient_to_mb
 
-@count_function
-def variant_count(row, cohort, filter_fn, normalized_per_mb, **kwargs):
-    patient_id = row["patient_id"]
-    return cohort.load_variants(
-        patients=[cohort.patient_from_id(patient_id)],
-        filter_fn=filter_fn,
-        **kwargs)
+def count_variants_function_builder(function_name, filterable_variant_function=None):
+    @count_function
+    def count(row, cohort, filter_fn, normalized_per_mb, **kwargs):
+        def count_filter_fn(filterable_variant, **kwargs):
+            assert filter_fn is not None, "filter_fn should never be None, but it is."
+            return ((filterable_variant_function(filterable_variant) if filterable_variant_function is not None else True) and
+                    filter_fn(filterable_variant, **kwargs))
+        patient_id = row["patient_id"]
+        return cohort.load_variants(
+            patients=[cohort.patient_from_id(patient_id)],
+            filter_fn=count_filter_fn,
+            **kwargs)
+    count.__name__ = function_name
+    return count
 
-@count_function
-def exonic_variant_count(row, cohort, filter_fn, normalized_per_mb, **kwargs):
-    def exonic_filter_fn(filterable_effect, **kwargs):
-        assert filter_fn is not None, "filter_fn should never be None, but it is."
-        return (isinstance(filterable_effect.effect, Exonic) and
-                filter_fn(filterable_effect, **kwargs))
-    # This only loads one effect per variant.
-    patient_id = row["patient_id"]
-    return cohort.load_effects(
-        only_nonsynonymous=False,
-        patients=[cohort.patient_from_id(patient_id)],
-        filter_fn=exonic_filter_fn,
-        **kwargs)
+def count_effects_function_builder(function_name, only_nonsynonymous, filterable_effect_function=None):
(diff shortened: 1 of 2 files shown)


cohorts/functions.py
```diff
@@ -68,150 +68,128 @@ def get_patient_to_mb(cohort):
...
-                filter_fn(filterable_effect, **kwargs))
-    # This only loads one effect per variant.
-    patient_id = row["patient_id"]
-    return cohort.load_effects(
-        only_nonsynonymous=True,
-        patients=[cohort.patient_from_id(patient_id)],
-        filter_fn=exonic_insertion_filter_fn,
-        **kwargs)
+def count_variants_function_builder(function_name, filterable_variant_function=None):
+    @count_function
```
jburos: This could benefit from a docstring

cohorts/functions.py
```diff
@@ -68,150 +68,128 @@ def get_patient_to_mb(cohort):
...
+                    filter_fn(filterable_variant, **kwargs))
+        patient_id = row["patient_id"]
+        return cohort.load_variants(
+            patients=[cohort.patient_from_id(patient_id)],
+            filter_fn=count_filter_fn,
+            **kwargs)
+    count.__name__ = function_name
+    return count
+
+def count_effects_function_builder(function_name, only_nonsynonymous, filterable_effect_function=None):
```
jburos: Same here - would be good to include a docstring

cohorts/functions.py
```diff
@@ -68,150 +68,128 @@ def get_patient_to_mb(cohort):
...
+            patients=[cohort.patient_from_id(patient_id)],
+            filter_fn=count_filter_fn,
+            **kwargs)
+    count.__name__ = function_name
+    return count
+
+variant_count = count_variants_function_builder("variant_count")
+
+snv_count = count_variants_function_builder(
+    "snv_count",
```
jburos, line 122: it would be nice to include an optional docstring for these generated functions. Perhaps you can even construct a generic string based on contents of `filterable_variant_function` ?

jburos (approved):
LGTM & makes sense ; I like the consistent naming. Thanks. Just a few minor suggestions but feel free to ignore.

cohorts/functions.py
tavinathanson: Seems non-trival to add a dynamic docstring; not sure it's worth it, esp. if the docstring can be inferred from the parameters?

jburos: most of it can be inferred from the parameters. One option would be to paste the code of the filter function if no docstring is provided? Otherwise would be pretty simple to accept an optional parameter `docstring` & pass that to `count_function.__doc__` if provided

tavinathanson: I didn't think `__doc__` was writable?

tavinathanson: Looks like it is, I'll add something!

tavinathanson:
@jburos added auto-generated docstrings, good to merge on 🍏 ?

jburos:
Yes! Looks good. Thanks Tavi

PR #189, opened 2017-02-24
Add a median VAF filter function and a filter_fn watermark

### LINKED ISSUE #183: Add watermark for default filter_fn on Cohort initialization
e.g. "Cohort is filtered by abc_filter by default"

### LINKED ISSUE #190: additional_data should automatically create Patient attributes
`p = Patient(..., additional_data={"hello": 1, "goodbye": 2})` should automatically populate `p.hello` and `p.goodbye`.

Perhaps the same for `Sample`s.

### LINKED ISSUE #191: Add frameshift count function
Count only frameshifts.

tavinathanson:
Moving some things from my `data.py` to `Cohorts`.

* Add a function for median VAF.
* Add a `filter_fn` watermark: fixes https://github.com/hammerlab/cohorts/issues/183.

Also in this PR:

* Automatically populate `Patient` object attributes based on `additional_data`: fixes https://github.com/hammerlab/cohorts/issues/190.
* Do the same with `Sample`.
* Add some more functions: fixes https://github.com/hammerlab/cohorts/issues/191.

### DIFF
--- a/cohorts/utils.py
+++ b/cohorts/utils.py
@@ -185,3 +185,10 @@ def get_logger(name, level=logging.INFO):
         logger.handlers = []
     logger.setLevel(level)
     return logger
+
+def set_attributes(obj, additional_data):
+    if additional_data is not None:
+        for key, value in additional_data.items():
+            if hasattr(obj, key):
+                raise ValueError("Key %s in additional_data already exists in this object" % key)
+            setattr(obj, _strip_column_name(key), value)
--- a/cohorts/utils.py
+++ b/cohorts/utils.py
@@ -187,8 +187,12 @@ def get_logger(name, level=logging.INFO):
     return logger
 
 def set_attributes(obj, additional_data):
-    if additional_data is not None:
-        for key, value in additional_data.items():
-            if hasattr(obj, key):
-                raise ValueError("Key %s in additional_data already exists in this object" % key)
-            setattr(obj, _strip_column_name(key), value)
+    """
+    Given an object and a dictionary, give the object new attributes from that dictionary.
+
+    Uses _strip_column_name to git rid of whitespace/uppercase/special characters.
+    """
+    for key, value in additional_data.items():
+        if hasattr(obj, key):
+            raise ValueError("Key %s in additional_data already exists in this object" % key)
+        setattr(obj, _strip_column_name(key), value)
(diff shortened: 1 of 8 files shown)


tavinathanson:
@jburos added some more stuff, feel free to review at this point.

cohorts/utils.py
```diff
@@ -185,3 +185,10 @@ def get_logger(name, level=logging.INFO):
         logger.handlers = []
     logger.setLevel(level)
     return logger
+
+def set_attributes(obj, additional_data):
+    if additional_data is not None:
+        for key, value in additional_data.items():
+            if hasattr(obj, key):
+                raise ValueError("Key %s in additional_data already exists in this object" % key)
```
jburos: Seems like you might run into this a fair amount, at least at the patient level. E.g. if the additional_data includes a variable like "hla_alleles" or "deceased" which has already been set. Is this the intended behavior?

jburos (approved):
In general LGTM - put a minor comment but isn't blocking. Lots of useful things in here. 

These could probably use some additional test cases (e.g. for median_vaf_purity & additional_data) but not that pressing.

cohorts/utils.py
tavinathanson: Yeah, I think that's a great point and one I was/am a little worried about.

I wanted to error to make sure that we didn't accidentally override important variables, and it doesn't seem too hard for the user to remove or rename the input `DataFrame` to work around the error.

I don't feel strongly, but one option is to just put this in and see what errors we hit?

cohorts/utils.py
```diff
@@ -185,3 +185,10 @@ def get_logger(name, level=logging.INFO):
...
+            setattr(obj, _strip_column_name(key), value)
```
tavinathanson: I should make it clear that this is happening in the docstring.

tavinathanson: @jburos I just updated `set_attributes` to remove the `None` check at the beginning, instead moving that to `Sample` and `Patient`. Also added a docstring.

jburos:
Nice - thanks @tavinathanson ! LGTM. I like the use of strip_column_name. If the conflict happens a lot we may want to catch the error & provide a more descriptive message in the init function for either Patient or Sample. But for now it's probably good as-is.

tavinathanson:
@jburos I had to fix a few failing tests due to e.g. `os` existing in `additional_data` as well as in `Patient` already. I'm fixing by removing that value from the `additional_data` dictionary, which actually seems correct based on the meaning of "additional data". But we'll definitely hit this error in our other cohorts. A simple replacement of e.g. `id=row["id"]` with `id=row.pop("id")` should fix things for all problematic columns in RCC/bladder/etc.

PR #199, opened 2017-04-28
Update encoding and str handling for Py2/3 compatibility

tavinathanson:
In this PR:

* Add `encoding="latin-1"` to `load_maf`; was needed for Python 3, which defaults to `utf-8`. (Though a bit confusing because Python 3 was working for me previously, but perhaps because the cache was older?)
* Use `string_types` rather than `str` in `require_id_str` to account for `str` and `unicode` in Python 2.
* Update requirements to a not-yet-real version of `varcode`.

Depends on https://github.com/hammerlab/varcode/pull/223


jburos (approved):
Thanks Tavi! also LGTM. should we wait for varcode to release & re-run tests before merging? I will create an issue for a `str(id)` unit test, though not highest priority to add in now

tavinathanson:
@jburos yeah, we should wait for `varcode` and re-run to get 🍏

PR #205, opened 2017-05-16
Update Pageant logic

tavinathanson:
Allow for single-sample or tumor/normal coverage, and update columns/etc. to the latest Pageant format. Allow the Pageant output file folder names to be specified as functions of the `Patient`.

### DIFF
--- a/cohorts/cohort.py
+++ b/cohorts/cohort.py
@@ -543,6 +543,7 @@ class Cohort(Collection):
             ## capture hash for any function within closure
             if inspect.isfunction(val):
                 closure.append(self._hash_filter_fn(val))
+        closure.sort() # Sorted for file name consistency
         closure_str = "null" if len(closure) == 0 else "-".join(closure)
         # construct final string comprising hashed components
         hashed_fn = ".".join(["-".join([filter_fn_name,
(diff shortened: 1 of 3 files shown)


jburos (approved):
LGTM - thanks tavi

tavinathanson:
@jburos weird investigation into a test failure led me to notice that pandas 0.20.x recently came out. I had trouble reproducing the test failure, but when I locally installed the latest pandas I did.

So then the question was: what's wrong with this?

```
     filter_mask = neoantigens_df.apply(
            lambda row: filter_fn(
                FilterableNeoantigen(neoantigen_row=row,
                                     variant_collection=variant_collection,
                                     patient=patient)),
            axis=1,
        return neoantigens_df[filter_mask]
```

As far as I can tell, what's wrong is that Pandas expects `filter_mask` to be a `Series`, but sometimes (namely when it's empty) it's a `DataFrame`. I added `reduce=True` to make it always return a `Series`. This feature has existed for a while AFAICT, so I'm unclear what broke in 0.20.x; but this seems to address the issue.

I'll file a new issue to write a test for neoantigens other than missing ones.

tavinathanson:
I looked at some Pandas code, and saw that some parts of `apply` were changed recently, though didn't find the culprit.

jburos:
nice find Tavi - makes me happy we have test code :) thanks for the description

tavinathanson:
@jburos https://github.com/hammerlab/cohorts/pull/205/commits/83d69ed8683c575db1847d718c14387db59d5640 addresses the additional test failures we saw; can you review the code?

jburos:
@tavinathanson looks great - thanks for digging into this!

tavinathanson:
@jburos good catch! Fixed.

PR #206, opened 2017-05-17
Add optional show_progress

tavinathanson:
Often, `tqdm` prints progress bars on multiple lines and it can be annoying to scroll through. @jburos any idea why?

Until we resolve that, perhaps we can just make `tqdm` optional? Hence this PR.

### DIFF
--- a/cohorts/cohort.py
+++ b/cohorts/cohort.py
@@ -75,6 +75,8 @@ class Cohort(Collection):
         A list of `Patient`s for this cohort.
     cache_dir : str
         Path to store cached results, e.g. cached variant effects.
+    show_progress : bool
+        Whether or not to show DataFrame application progress as an increasing percentage.
     kallisto_ensembl_version : int
         Cached release version to use from pyensembl to annotate Kallisto data
     cache_results : bool
@@ -119,6 +121,7 @@ class Cohort(Collection):
     def __init__(self,
                  patients,
                  cache_dir,
+                 show_progress=True,
                  kallisto_ensembl_version=None,
                  cache_results=True,
                  extra_df_loaders=[],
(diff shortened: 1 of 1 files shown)


jburos:
@tavinathanson agreed - i've noticed that too. When I was testing previously, this happened when logger tried to write to stdout or other status things used `print` instead of logger for example. Any of these things that write to stdout directly break compatibility with tqdm. It might makes sense to create an issue for this - #207 

Re: implementation, we could probably leave this even if `tqdm` printing is fixed, since this would also allow us to remove the strict dependency on `tqdm`. What do you think about removing this from `requirements.txt` & defaulting `show_progress` to False if `tqdm` is not installed?

I noticed this was done for `rpy2` in [PyVCF](https://github.com/jamescasbon/PyVCF/blob/master/vcf/filters.py#L1): 

```py
try:
    from rpy2 import robjects
except:
    robjects = None
```

tavinathanson:
@jburos cool, thanks for filing that issue. I suppose the downside of not requiring `tqdm` would be that people wouldn't automatically install it on a `pip install cohorts` and may not know that they're missing it.

Not sure who we're focused on helping with not requiring `tqdm` at the moment? Maybe worth tabling, and avoiding the slight increase in complexity, until/unless anyone has issues with that?

jburos:
@tavinathanson was just thinking the fewer requirements the better, overall. OK if you want to table. in PyVCF, the user sees a message if they try to use a feature that requires rpy2 but it isn't installed, so that they know to install it. But admittedly, rpy2 is messier than tqdm.

jburos:
Otherwise, LGTM. OK to merge!

tavinathanson:
@jburos I generally agree with the fewer requirements the better, though it feels a little unnecessary at this point given our low number of requirements.

Happy to keep the conversation going, but will merge for now.

PR #208, opened 2017-05-18
Require Python 3

tavinathanson:
No longer support Python 2, including:
* No longer test with Python 2.
* No longer use compatibility layers for Py2/3.
* Remove version checking logic.
* Clarify on README.
* Add `python_requires` constraint in `setup.py`.

Also update 2016 to 2017 in files.

I figured we should go all out and not allow Python 2 usage at all if we're no longer testing with Python 2. Otherwise, issues like `str` vs. `unicode` resulting in the wrong number of patients in the `Cohort` will be highly likely.

### DIFF
--- a/cohorts/cohort.py
+++ b/cohorts/cohort.py
@@ -23,7 +23,6 @@ from copy import copy
 import dill
 import hashlib
 import inspect
-import sys
 import logging
 
 # pylint doesn't like this line
@@ -573,9 +572,6 @@ class Cohort(Collection):
             filter_fn_name = self._get_function_name(filter_fn)
             logger.debug("loading variants for patient {} with filter_fn {}".format(patient.id, filter_fn_name))
             use_filtered_cache = use_cache
-            if sys.version_info < (3, 3):
-                logger.info("... disabling filtered cache due to python version")
-                use_filtered_cache = False
 
         ## confirm that we can get cache-name (else don't use filtered cache)
         if use_filtered_cache:
--- a/cohorts/cohort.py
+++ b/cohorts/cohort.py
@@ -24,6 +24,7 @@ import dill
 import hashlib
 import inspect
 import logging
+import pickle
 
 # pylint doesn't like this line
 # pylint: disable=no-name-in-module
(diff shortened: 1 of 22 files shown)


cohorts/cohort.py
```diff
@@ -576,9 +573,6 @@ def _load_single_patient_variants(self, patient, filter_fn, use_cache=True, **kw
             filter_fn_name = self._get_function_name(filter_fn)
             logger.debug("loading variants for patient {} with filter_fn {}".format(patient.id, filter_fn_name))
             use_filtered_cache = use_cache
-            if sys.version_info < (3, 3):
```
jburos, line 579: Given our requirement is for python >=3 we should probably still check for versions 3.1 and 3.2?

jburos: and/or, add these versions to our test matrix?

setup.py
```diff
@@ -67,6 +67,7 @@
         ],
         install_requires=install_requires,
         dependency_links=dependency_links,
+        python_requires=">=3",
```
jburos: See above comment re: dependency on version >= 3.3

jburos (changes requested):
Some minor comments but looks good to me - thanks Tavi! excited to make a clean break

cohorts/cohort.py
tavinathanson: Added 3.3, 3.4 and 3.6 to our matrix, and requiring >= 3.3 to use Cohorts.

tavinathanson:
@jburos I think I addressed your comments, lmk!

jburos (approved):

tavinathanson:
@jburos turns out `pysam` `0.9.0`, which we `conda install` in Travis (https://github.com/hammerlab/cohorts/pull/49#issuecomment-225646603), requires 3.4.

So, instead, I made it `>= 0.9.0` and also added `pysam` to our `requirements.txt` (since we do directly use it now). I also removed `3.3` from Travis, since no current version of `pysam` on `Conda` appears to support `3.3`.

Hopefully that'll all work.

PR #210, opened 2017-05-19
Add new PyPi server

tavinathanson:
We now get this message when trying to deploy to PyPi:

https://github.com/pybuilder/pybuilder/issues/453

Will mention to the rest of the lab if this fix works.


jburos:
Looks good to me Tavi - let's see if it works!

Issue #214, opened 2017-06-05
Add `cache_root_dir` attribute


jburos:
When analyzing multiple cohorts in the context of a single paper, we often want to be able to group the cache-dir for each cohort so that they can be copied, deleted, or referenced collectively.

Some use cases include:

- if we are versioning the caches for a given analysis. e.g. we might have one cache with & one without certain settings (e.g. BQSR) for somatic mutation calling.
- or, if we want to "wipe" or force-refresh the cache(s) for particular analysis.

### Rationale for cache_root_dir

The simplest version of this functionality (until a more robust cache-file management solution is implemented - see #175) is currently to include an optional `cache_root_dir` attribute in the Cohort object. 

The idea is that related Cohort objects would store their caches within the same `cache_root_dir`.

**note**: This is optional in that cohorts created without `cache_root_dir`, **or** that are initiated with an absolute path for the `cache_dir` should _completely ignore the input for `cache_root_dir`. 

Specifically, in both scenarios: 

- `cache_door_dir` attribute set to None, and
- `cache_dir` set to the value of cache_dir used to create the Cohort

### Also: standard cache-dir naming 

As a secondary feature, it would be helpful to be able to name the cache-dirs for each cohort according to a standard protocol. 

This would either support 

- a consistent naming for each cohort's cache_dir _within_ the `cache_root_dir`
- and/or allow for the cache-dir to depend on input parameters.

The second feature is implemented by the optional `cache_dir_kwargs` parameter 

For example, say we have a function that returns a TCGA `Cohort` containing TCGA data, where the composition of the cohort depends on an input attribute "tumor_type". In this case, we might want a different cache-dir for each tumor-type.

tavinathanson:
Given that the `Cohort` object is the highest level that `cohorts` knows about, I feel like we'd want to introduce knowledge of multiple cohorts into the library before adding `cache_root_dir`?

tavinathanson:
I see that you already have a PR; suppose it's reasonable to start with that and go from there 👍

jburos:
yeah - ok. Am open to other implementations, but also wanted to write down thought process behind that PR for now

PR #216, opened 2017-06-05
add cache_root_dir attribute & behavior

jburos:
Adding cache_root_dir per #214

### DIFF
--- a/cohorts/utils.py
+++ b/cohorts/utils.py
@@ -17,6 +17,25 @@ import warnings
 from collections import namedtuple
 import sys
 import logging
+from os import path
+
+def get_cache_dir(cache_dir, cache_root_dir=None, *args, **kwargs):
+    """
+    Return full cache_dir, according to following logic:
+        - if cache_dir is a full path (per path.isabs), return that value
+        - if not and if cache_root_dir is not None, join two paths
+        - otherwise, log warnings and return None
+    Separately, if args or kwargs are given, format cache_dir using kwargs
+    """
+    cache_dir = cache_dir.format(*args, **kwargs)
+    if path.isabs(cache_dir):
+        return(cache_dir)
+    if cache_root_dir is not None:
+        return path.join(cache_root_dir, cache_dir)
+    else:
+        logger.warning("cache dir is not full path & cache_root_dir not given. Caching may not work as expected!")
+    return(None)
+
 
 class DataFrameHolder(namedtuple("DataFrameHolder", ["cols", "df"])):
     """Holds a DataFrame along with associated columns of interest."""
--- a/cohorts/utils.py
+++ b/cohorts/utils.py
@@ -19,6 +19,8 @@ import sys
 import logging
 from os import path
 
+logger = get_logger(__name__, level=logging.INFO)
+
 def get_cache_dir(cache_dir, cache_root_dir=None, *args, **kwargs):
     """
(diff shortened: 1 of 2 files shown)


cohorts/utils.py
```diff
@@ -17,6 +17,25 @@
...
+def get_cache_dir(cache_dir, cache_root_dir=None, *args, **kwargs):
+    """
+    Return full cache_dir, according to following logic:
+        - if cache_dir is a full path (per path.isabs), return that value
+        - if not and if cache_root_dir is not None, join two paths
+        - otherwise, log warnings and return None
+    Separately, if args or kwargs are given, format cache_dir using kwargs
+    """
+    cache_dir = cache_dir.format(*args, **kwargs)
+    if path.isabs(cache_dir):
```
tavinathanson, line 33: This logic feels a little incomplete to me. What if an absolute path is passed in along with a cache_root_dir? Can we have a warning for that scenario?

cohorts/utils.py
```diff
@@ -17,6 +17,25 @@
...
+        return(cache_dir)
+    if cache_root_dir is not None:
+        return path.join(cache_root_dir, cache_dir)
+    else:
+        logger.warning("cache dir is not full path & cache_root_dir not given. Caching may not work as expected!")
+    return(None)
```
tavinathanson: Should be: `return None`

cohorts/utils.py
```diff
@@ -17,6 +17,25 @@
...
+    """
+    Return full cache_dir, according to following logic:
+        - if cache_dir is a full path (per path.isabs), return that value
+        - if not and if cache_root_dir is not None, join two paths
+        - otherwise, log warnings and return None
+    Separately, if args or kwargs are given, format cache_dir using kwargs
+    """
+    cache_dir = cache_dir.format(*args, **kwargs)
+    if path.isabs(cache_dir):
+        return(cache_dir)
```
tavinathanson: Should be: `return cache_dir`

tavinathanson (commented):
Can you give me an example of how the format string is used?

cohorts/utils.py
jburos: Sure, I can add a warning. 

I'm thinking this would most likely occur if there's a default `cache_root_dir` (in some cohort's version of what we call `data.py`) & the user then chooses to override the parameter for `cache_dir`. In this case, we would want to effectively ignore the root dir & otherwise handle the cache_dir. 

Also technically if you evaluate an absolute path relative to another path, you just get the absolute path out.

jburos:
I use the format-string in our [tcga_data.py](https://github.com/hammerlab/msk-bms-lung/blob/tcga-thresholds/msk_bms_lung/tcga_data.py) within the lung-cohort, so that the cache-dir is named according to the tumor type (which I think we call "disease": `luad` vs `lusc`). 

For example, in that file we have:

```py
def load_tcga(disease, legacy_bams=True, cache_dir="cache-new-tcga-{disease}", ... ):
  ... 
  cache_dir = get_cache_dir(cache_dir=cache_dir, cache_root_dir=cache_root_dir, disease=disease)

```

In that case, I computed the cache-dir outside of the cohort & passed it in; this is an attempt to support similar functionality within the cohort object.

Let me know what you think & if I should add more detail on this to issue #214.

tavinathanson:
👍  LGTM

Issue #221, opened 2017-07-06
Invalid RGBA argument error using `plot_survival`


jburos:
I'm using `plot_survival` on a cohort & am getting a strange error: 

**Input:**
```python
my_cohort.plot_survival(on='age')
```

**Output:**

```python
ValueError: Invalid RGBA argument: 0.6627450980392157
```

Using Python 3.5.2 & matplotlib 2.0.2 (installed via conda)
```sh
Python 3.5.2 :: Continuum Analytics, Inc.
matplotlib                2.0.2               np113py35_0
```

Here is the traceback:

```python
# no condition 29
# with condition 27
Out[14]:
<lifelines.StatisticalResult: 
Results
   null distribution: chi squared
   alpha: 0.95
   df: 1
   test: logrank
   t 0: -1

   __ p-value ___|__ test statistic __|____ test result ____|__ is significant __
         0.69421 |              0.155 |  Cannot Reject Null |       False       
>
/home/jacquelineburos/miniconda3/envs/rcc/lib/python3.5/site-packages/matplotlib/font_manager.py:1297: UserWarning: findfont: Font family ['sans-serif'] not found. Falling back to DejaVu Sans
  (prop.get_family(), self.defaultFamily[fontext]))
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
/home/jacquelineburos/miniconda3/envs/rcc/lib/python3.5/site-packages/matplotlib/colors.py in to_rgba(c, alpha)
    140     try:
--> 141         rgba = _colors_full_map.cache[c, alpha]
    142     except (KeyError, TypeError):  # Not in cache, or unhashable.

KeyError: (0.6627450980392157, None)

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
/home/jacquelineburos/miniconda3/envs/rcc/lib/python3.5/site-packages/IPython/core/formatters.py in __call__(self, obj)
    305                 pass
    306             else:
--> 307                 return printer(obj)
    308             # Finally look for special method names
    309             method = get_real_method(obj, self.print_method)

/home/jacquelineburos/miniconda3/envs/rcc/lib/python3.5/site-packages/IPython/core/pylabtools.py in <lambda>(fig)
    225 
    226     if 'png' in formats:
--> 227         png_formatter.for_type(Figure, lambda fig: print_figure(fig, 'png', **kwargs))
    228     if 'retina' in formats or 'png2x' in formats:
[...]

PR #223, opened 2017-07-06
Fix RGBA error with plot_survival

jburos:
Address issue #221, by which an error is raised on lifelines version >= 0.9.3.2.  Though, to be fair, it's not clear if the error is due to matplotlib or lifelines. 

We will fix the error for those versions of lifelines & incrementing the requirements to depend on more recent versions of lifelines.


requirements.txt
```diff
@@ -5,7 +5,7 @@ topiary>=0.1.0, <0.2.0
 mhctools>=0.3.0, <0.4.0
 varcode>=0.5.15, <0.6.0
 pyensembl>=1.0.1, <1.1.0
-lifelines>=0.9.1.0
+lifelines>=0.9.3.2
```
armish, line 8: Just OOC: is there a particular reason as to why we are way behind their release cycle? Looks like their most recent release is 0.11.1 which actually seems to address a similar issue: https://github.com/CamDavidsonPilon/lifelines/issues/304

armish (approved):
Looks good to me, @jburos. I hate it when such conflicts arise just out of nowhere but glad this had an easy fix.

jburos:
This is a good point - increasing the version to 0.10+ would also fix this. Though in general I try to keep requirements as flexible as possible in case of conflicts. What do you think, would it be cleaner to force a more recent version?

PR #224, opened 2017-07-07
Feature - plot survival by category

jburos:
building off earlier PRs, this PR adds support for a categorical variable in `plot_survival`.

Now, if two groups are provided, the result is plotted as if it were a float (cut at median or threshold) or boolean value:

screen shot 2017-07-06 at 8 10 03 pm

If the value contains data for more than 2 groups, a CoxPH model is fit & survival curves for each group are plotted:
screen shot 2017-07-06 at 8 13 03 pm


jburos:
Oops realize now I need to cherry pick these commits to avoid the warn-on-vcf-error hack

Issue #217, opened 2017-06-15
support plotting survival curves for more than 2 groups


jburos:
Came up in a recent conversation on slack; and, have personally wanted this feature for a while.

Some scenarios that came up in my experience:

- plot survival curves by a string variable or category, IE "pd-l1".  
- user provides a function to map rows to groups, where the resulting groups number more than 2.
- user plots on a variable ("missense_snv_count") but provides more than 1 threshold value.  

Copying convo from slack here for context:

@timodonnell : 
> I’d like to generate a survival curve for various levels of T cell infiltration

@tavinathanson:
> how many curves in the plot? `cohorts` right now allows just 2 curves based on a binary feature or e.g. split by median value

jburos:
Linking to related PR: #225. This PR implements the first two scenarios of the 3 listed above.

PR #225, opened 2017-07-07
Feature plot_survival by category (clean branch)

jburos:
This is a re-submit of a previous PR #224. This adds basic support for plotting survival according to values of a string (categorical) variable, using `plot_survival`. See related issue #217.

As before, if you use a float or boolean value, the survival curves are plotted for True/False groups, or above/below a threshold (which defaults to a median value). 

This PR adds support to be able to plot by a string variable -- after casting that string to a category. 

As with the existing behavior with float/boolean values, if two groups are provided, the survival curves are plotted for the two groups & a log-rank test result is reported. 

screen shot 2017-07-06 at 8 10 03 pm

However, if the value contains data for more than 2 groups, a CoxPH model is fit & more than two survival curves are plotted:

screen shot 2017-07-06 at 8 13 03 pm

There are other scenarios in #217 which are not covered here (ie passing in more than one threshold value, etc), and I can imagine more extensions being useful (e.g. fitting a cox-ph model to a float value but plotting by the median). But for now this covers the majority of use cases.


cohorts/survival.py
```diff
@@ -72,59 +77,89 @@ def plot_kmf(df,
...
             label_suffix += " (median)"
         default_label_with_condition = "%s > %s" % (condition_col, label_suffix)
-    else:
+        with_condition_label = with_condition_label or default_label_with_condition
+        no_condition_label = no_condition_label or default_label_no_condition
+        label_map = {False: no_condition_label,
+                     True: with_condition_label}
+        color_map = {False: no_condition_color,
+                     True: with_condition_color}
+    elif df[condition_col].dtype == 'O' or df[condition_col].dtype.name == "category":
```
armish: Looks like a `lambda`-based solution would help cover the last two use cases in a similar way the other plotting utilities work; but if we don't immediately need it, I would say just leave that for another day/PR since the `threshold` argument might potentially complicate the implementation :man_shrugging:

cohorts/survival.py
```diff
@@ -72,59 +77,89 @@ def plot_kmf(df,
...
+        condition = df[condition_col].astype("category")
+        label_map = dict()
+        [label_map.update({condition_value: '{} = {}'.format(condition_col,
+                                                    condition_value)})
+                 for condition_value in condition.unique()]
+        rgb_values = sb.color_palette("Set2", len(label_map.keys()))
```
armish: Looking at your example in the comments, the colors are pretty distinct for 2 categories but they blend in with the background when there are more of them. Maybe making them a bit darker could help with that?

Feel free to ignore this nit-pick if you think they already look great :)

armish (approved):
Looks great, @jburos! Thanks for tackling this.

cohorts/survival.py
jburos: Actually, the `lambda` solution does work already, b/c tavi's `cols` logic is processed before this call to plot_kmf. Previously your lambda had to return a boolean but now can return a string or multiple categories. 

One thing I'd like to do is refactor this to not have the `no_condition_*`, `with_condition_*` etc. and instead take the color_map & label_map parameters directly. I wrote the function to support this but will implement it another day.

jburos: E.g.: 

screen shot 2017-07-07 at 12 28 23 pm

cohorts/survival.py
jburos: This is a good point - thanks. I didn't give this much thought because I was planning to allow the user to pick a "palette". Here I ended up hard-coding one, for sake of time, but I think I can allow the user to pass one in.

jburos: I added an option to change the color_palette to something with more contrast. `cohort.plot_survival(on='bor', color_palette="Set1")`

cohorts/survival.py
armish: Ah, sorry my bad - didn't see that logic in the collapsed view 🙏 

Cohorts is upping its game :)

Issue #175, opened 2017-01-12
Simplify cohorts caching


tavinathanson:
We use the caching functions a lot, but don't expose any simple primitives to make caching cleaner and less repetitive.

tavinathanson:
Continues to be needed!

jburos:
linking to a related issue #248

PR #230, opened 2017-07-17
Update warning and associated comment

tavinathanson:
Add a bit of clarity w.r.t. the `StopIteration` catch.


jburos (approved):
Looks good - thanks @tavinathanson. One thing I'll point out is, once your variants are cached this warning never shows up again. Something maybe to consider when we get to #175

tavinathanson:
@jburos yeah, we had an issue like that a while ago, where we accidentally cached something when we hit an error and then it was just empty with no error after that. At least, here, we should generally only be doing that when it's an empty VCF.

That said, it does feels somewhat error-prone. I'll file an additional issue about that.

Issue #228, opened 2017-07-17
Error using median_vaf_purity on a combined-polyphen-snpeff.vcf


jburos:
Getting the following error when trying to estimate `median_vaf_purity` on a combined/merged VCF file.

```python
/home/jacquelineburos/projects/cohorts/cohorts/functions.py in grab_vaf(variant)
    271     def grab_vaf(variant):
    272         filterable_variant = FilterableVariant(variant, variants, patient)
--> 273         return variant_stats_from_variant(variant, filterable_variant.variant_metadata).tumor_stats.variant_allele_frequency
    274     vafs = [grab_vaf(variant) for variant in variants]
    275     return 2 * pd.Series(vafs).median()

/home/jacquelineburos/projects/cohorts/cohorts/variant_stats.py in variant_stats_from_variant(variant, metadata, merge_fn)
    189             stats = mutect_somatic_variant_stats(variant, variant_metadata)
    190         else:
--> 191             raise ValueError("Cannot parse sample fields, variant file {} is from an unsupported caller.".format(variant_file))
    192         all_stats.append(stats)
    193     return merge_fn(all_stats)

ValueError: ('Cannot parse sample fields, variant file /mnt/disks/data/xxxx-1normals-1tumors--b37decoy/passed_somatic_combined_polyphen-snpeff.vcf is from an unsupported caller.', 'occurred at index 0')
```

tavinathanson:
@jburos it currently looks for "mutect" or "strelka" in the filename to figure out how to parse the file. Silly, I know. We need to do better.

jburos:
@tavinathanson yep, i figured it was something like this. just noting for now, it's not urgent atm. seems like having an "other" category & parsing like a generic vcf would be useful -- i assume there _is_ a generic vcf format?

tavinathanson:
@jburos my understanding is that we can't rely on a standard format for things like depth and VAF, hence needing to understand the caller in order to grab them. But we should be able to infer the format.

armish:
Here is the backstory on this one: Strelka's VCF implementation is not complete and VCF validators fail them because of the weird choice of fields/annotations they use in the VCF. The funny things is that its developers are aware of this and the whole Biostars has been complaining about this, but still no change/fix.

[...]

PR #232, opened 2017-07-18
"guess" at variant file format, if not obvious from filename

jburos:
Quick hack to get around errors in #228 .. tried to infer file types from meta-data, but turned out to be too inconsistently formatting, surprisingly even among results from the same pipeline. 

At any rate, this appears to work for now. With a sample size of one strelka & one mutect file, it appears that reading one file type as if it were the other fails in both directions.


cohorts/variant_stats.py
```diff
@@ -158,6 +160,30 @@ def maf_somatic_variant_stats(variant, variant_metadata):
...
+    """Retrun True if variant_file given is in .maf format
+    """
+    return ".maf" in variant_file.lower()
+
+def _vcf_is_mutect(variant_file):
+    """Return True if variant_file give is in mutect format
+    """
+    if "mutect" in variant_file.lower():
+        return True
+    else:
```
tavinathanson: I think this would be safer if you just checked for the presence of the thing and returned `False` otherwise, without try/catch; I don't like catching errors so broadly, as it seems like it'd be easy to hide e.g. a file reading issue.

cohorts/variant_stats.py
```diff
@@ -158,6 +160,30 @@ def maf_somatic_variant_stats(variant, variant_metadata):
...
+        try:
+            vcf_reader = vcf.Reader(open(variant_file, "r"))
```
tavinathanson: Why do you need to open the VCF vs. use the `variant_metadata` already extracted by `varcode`?

cohorts/variant_stats.py
```diff
@@ -181,13 +207,19 @@ def variant_stats_from_variant(variant,
...
             stats = maf_somatic_variant_stats(variant, variant_metadata)
-        elif "strelka" in variant_file.lower():
+        elif _vcf_is_strelka(variant_file=variant_file):
             stats = strelka_somatic_variant_stats(variant, variant_metadata)
-        elif "mutect" in variant_file.lower():
+        elif _vcf_is_mutect(variant_file=variant_file):
             stats = mutect_somatic_variant_stats(variant, variant_metadata)
         else:
-            raise ValueError("Cannot parse sample fields, variant file {} is from an unsupported caller.".format(variant_file))
+            try:
```
tavinathanson: Here, too, my preference is checking for a particular Strelka field and a particular Mutect field, and avoiding the try/catches here, to avoid hiding other possible errors. Thoughts?

tavinathanson (changes requested):
Cool! Just a few comments.

cohorts/variant_stats.py
jburos: This is a good point. The error i'm trying to catch here is one caused by the meta-data being a different structure depending on the file type. I am about to push an edit that will try to isolate it better. 

Curious if you know of a better way to handle this ? ie Confirming that the item is a list of dicts, and that this dict contains a certain key seems like overkill.

cohorts/variant_stats.py
jburos: I think the underlying function should throw a specific error type we could catch, in order to avoid this. In general it seems having a greater diversity of specific errors would help avoid this.

cohorts/variant_stats.py
jburos: These are different. the variant_metadata is not the same thing as the file_metadata, afaict.

E.g.:
```python
(Pdb) variant_metadata.keys()
dict_keys(['id', 'filter', 'qual', 'info', 'sample_info', 'alt_allele_index'])
(Pdb) vcf_reader = vcf.Reader(open(variant_file, "r"))
(Pdb) vcf_reader.metadata.keys()
odict_keys(['fileformat', 'GATKCommandLine.CombineVariants', 'cmdline', 'content', 'fileDate', 'germlineIndelTheta', 'germlineSnvTheta', 'priorSomaticIndelRate', 'priorSomaticSnvRate', 'reference', 'source', 'source_version', 'startTime', 'SnpEffVersion', 'SnpEffCmd'])
```
Oddly, this dict contains very different things even for files generated from the same epidisco pipeline. 

for example, in this case the vcf has a metadata field which is particularly helpful:

```python
(Pdb) vcf_reader.metadata['content']
['strelka somatic indel calls']
```

I didn't look for particular fields based mostly on arman's feedback, which suggested that the set of fields may be different for different versions within a particular format. Also trying to read the file according to mutect or strelka format would implicitly test for presence of fields. So this is the fall-back mechanism in a way.

cohorts/variant_stats.py
```diff
@@ -158,6 +161,45 @@ def maf_somatic_variant_stats(variant, variant_metadata):
         normal_stats = _maf_variant_stats(variant, variant_metadata, prefix="n")
     return SomaticVariantStats(tumor_stats=tumor_stats, normal_stats=normal_stats)
 
+def _vcf_is_strelka(variant_file, variant_metadata):
```
tavinathanson, line 163: For consistency, can you replace ' with " in this PR?

cohorts/variant_stats.py
```diff
@@ -158,6 +161,45 @@ def maf_somatic_variant_stats(variant, variant_metadata):
...
+        return True
+    elif 'NORMAL' in variant_metadata['sample_info'].keys():
+        return True
+    else:
+        vcf_reader = vcf.Reader(open(variant_file, "r"))
+        try:
+            vcf_type = vcf_reader.metadata['content']
+        except KeyError:
+            vcf_type = ''
+        if re.search(string=vcf_type.lower(), pattern='strelka'):
```
tavinathanson: Nice! Why not just `"strelka" in vcf_type.lower()` like above? Same for Mutect?

tavinathanson (commented):
Overall looks good, just minor comments! Maybe would be more ideal to add functionality to `varcode` when it imports `vcf` to save additional metadata, but this seems fine for now?

cohorts/variant_stats.py
jburos: sure. didn't know you could do that - thanks

cohorts/variant_stats.py
jburos: Sure :) I usually do search & replace, but this file has a few pre-existing `'` - so will update the few stragglers.

tavinathanson (approved):

PR #236, opened 2017-07-18
allow median-vaf-purity with patients having no variants

jburos:
Small fix so that `median_vaf_purity` doesn't error out when analyzing a cohort where only a subset of the patients have variants (or, some actually have zero variants).


tavinathanson (approved):
LGTM!

PR #237, opened 2017-07-18
Plot surv by strata

jburos:
Minor feature added to `plot_survival`, to support a `strata` keyword.

At the moment, this results in a separate survival plot or analysis being produced, one for each stratifying group. 

E.g.: 

3538a6dce4da4be7a91a099012aa46c7 _screen 20shot 202017-07-18 20at 202 25 35 20pm

(note that I have greyed out the sample names although in reality they printed as text).

### DIFF
--- a/cohorts/cohort.py
+++ b/cohorts/cohort.py
@@ -1320,7 +1320,7 @@ class Cohort(Collection):
         survival_units : str
             Unit of time for the survival measure, i.e. Days or Months
         strata : str
-            column name of stratifying variable
+            (optional) column name of stratifying variable
         ci_show : bool
             Display the confidence interval around the survival curve
         threshold : int or "median", optional
@@ -1340,57 +1340,25 @@ class Cohort(Collection):
             default_threshold = None
         else:
             default_threshold = "median"
-        if strata is None:
-            results = plot_kmf(
-                df=df,
-                condition_col=plot_col,
-                xlabel=survival_units,
-                ylabel="Overall Survival (%)" if how == "os" else "Progression-Free Survival (%)",
-                censor_col="deceased" if how == "os" else "progressed_or_deceased",
-                survival_col=how,
-                threshold=threshold if threshold is not None else default_threshold,
-                ax=ax,
-                ci_show=ci_show,
-                with_condition_color=with_condition_color,
-                no_condition_color=no_condition_color,
-                with_condition_label=with_condition_label,
-                no_condition_label=no_condition_label,
-                color_palette=color_palette,
-                label_map=label_map,
-                color_map=color_map,
-            )
-        else:
-            results = dict()
-            n_strata = len(df[strata].unique())
-            if ax is None:
-                f, ax = plt.subplots(n_strata, sharex=True)
-            a = 0
-            for strat, strat_df in df.groupby(strata):
-                if n_strata == 1:
-                    my_axis = ax
-                else:
-                    my_axis = ax[a]
-                title = "{}: {}".format(strata, strat)
-                print(title)
-                results[title] = plot_kmf(df=strat_df,
-                                        condition_col=plot_col,
-                                        xlabel=survival_units,
-                                        ylabel="Overall Survival (%)" if how == "os" else "Progression-Free Survival (%)",
-                                        censor_col="deceased" if how == "os" else "progressed_or_deceased",
-                                        survival_col=how,
-                                        threshold=threshold if threshold is not None else default_threshold,
-                                        ax=my_axis,
-                                        ci_show=ci_show,
-                                        with_condition_color=with_condition_color,
-                                        no_condition_color=no_condition_color,
-                                        with_condition_label=with_condition_label,
-                                        no_condition_label=no_condition_label,
-                                        color_palette=color_palette,
-                                        label_map=label_map,
-                                        color_map=color_map,
-                                        )
-                my_axis.set_title(title)
-                a += 1
+        results = plot_kmf(
+            df=df,
+            condition_col=plot_col,
+            xlabel=survival_units,
+            ylabel="Overall Survival (%)" if how == "os" else "Progression-Free Survival (%)",
+            censor_col="deceased" if how == "os" else "progressed_or_deceased",
+            survival_col=how,
+            strata_col=strata,
+            threshold=threshold if threshold is not None else default_threshold,
+            ax=ax,
+            ci_show=ci_show,
+            with_condition_color=with_condition_color,
+            no_condition_color=no_condition_color,
+            with_condition_label=with_condition_label,
+            no_condition_label=no_condition_label,
+            color_palette=color_palette,
+            label_map=label_map,
+            color_map=color_map,
+        )
         return results
 
     def plot_correlation(self, on, x_col=None, plot_type="jointplot", stat_func=pearsonr, show_stat_func=True, plot_kwargs={}, **kwargs):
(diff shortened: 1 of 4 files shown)


cohorts/cohort.py
```diff
@@ -1336,24 +1340,57 @@ def plot_survival(self,
...
-            ci_show=ci_show,
-            with_condition_color=with_condition_color,
-            no_condition_color=no_condition_color,
-            with_condition_label=with_condition_label,
-            no_condition_label=no_condition_label,
-            color_palette=color_palette,
-            label_map=label_map,
-            color_map=color_map,
-        )
+        if strata is None:
```
tavinathanson: Can you find a way to remove this chunk of code entirely and just have the loop? Too much copy pasting for my comfort :)

tavinathanson (changes requested):
LGTM! One minor comment

cohorts/cohort.py
jburos: i agree - at the time just needed to get this to work. i had an alternate approach in mind, will give it a try.

jburos: Updated this section to move strata functionality within the plot_kmf function. Still a bit cludgy (and still not a true stratified analysis), but LMK what you think.

cohorts/survival.py
```diff
@@ -76,6 +80,38 @@ def plot_kmf(df,
...
+        # create axis / subplots
+        if ax is not None:
+            raise ValueError('ax not supported with stratified analysis.')
+        n_strata = len(df[strata_col].unique())
+        f, ax = plt.subplots(n_strata, sharex=True)
+        a = 0 ## index for ax
+        # create results dict to hold per-strata results
+        results = dict()
+        # update strata_col to be None for each sub-call
+        arglist['strata_col'] = None
```
tavinathanson: This feels a bit confusing/hard to read, since `strata_col` isn't really `None`.

How about a function called `plot_kmf_single` that basically comprises everything after `# if strata is None, produce plot`?

Otherwise looks great!

cohorts/survival.py
```diff
@@ -76,6 +80,38 @@ def plot_kmf(df,
         print_as_title: bool, optional, whether or not to print text
           within the plot's title vs. stdout, default False
     """
+    # if strata not None, call plot_kmf for each of the strata
+    if strata_col:
+        # want this at the top of the function!
+        arglist = locals() ## save input args for reuse
```
tavinathanson: Using `locals()` feels a little hackish and unnecessary and easy to screw up of someone moves it after the first line; I'd prefer just passing the parameters through.

Also, I think it's probably not good to modify the `locals()` `dict` as you're doing with `arglist`?

https://docs.python.org/3/library/functions.html#locals
> modifications to the default locals dictionary should not be attempted.

cohorts/survival.py
```diff
@@ -76,6 +80,38 @@ def plot_kmf(df,
...
+            raise ValueError('ax not supported with stratified analysis.')
+        n_strata = len(df[strata_col].unique())
+        f, ax = plt.subplots(n_strata, sharex=True)
+        a = 0 ## index for ax
+        # create results dict to hold per-strata results
+        results = dict()
+        # update strata_col to be None for each sub-call
+        arglist['strata_col'] = None
+        # call plot_kmf for each strata, using same args
+        for strat_name, strat_df in df.groupby(strata_col):
```
tavinathanson: You could do `for a, (strat_name, strat_df) in enumerate(df.groupby(strata_col)):` to avoid declaring and incrementing `a`, but up to you if that's less clear.

Maybe call it `i`?

tavinathanson (commented):
A few more minor comments; if you disagree, feel free to ignore

armish:
@jburos: this is completely off-topic but if you hate those `matplotlib` error messages as much as I do, just shutdown your jupyters (or anything that makes use of `matplotlib`) and `rm -f ~/.cache/matplotlib/fontList.*`, which should get rid of those messages once you re-import the `matplotlib` once and for all.

cohorts/survival.py
jburos: cool - that seems useful. thanks

cohorts/survival.py
jburos: Yeah, I was trying to avoid copying & pasting all the parameters that many times .. feels a bit repetitive but may be preferable ... Your point about modifying the locals() dict is really spot on, .. I forget that this copies by reference not value. It general it seems this isn't a common pattern in Python, so will avoid it!

cohorts/survival.py
jburos: just FYI - not urgent - but i refactored these functions. Curious to know what you think.

cohorts/survival.py
```diff
@@ -18,69 +18,44 @@
...
-        label_map: dict, optional, mapping of labels to condition text.
-          defaults to "condition_name = condition_value", or *_condition_label
-          options in case of boolean operators.
-        color_palette: str, optional, name of sb.color_palette to use
-          if color_map not provided.
-        print_as_title: bool, optional, whether or not to print text
-          within the plot's title vs. stdout, default False
+    All inputs are required - this function is intended to be called by `plot_kmf`.
     """
+    # make color in puts consistent hex format
```
tavinathanson: Typo: `inputs`?

jburos:
@tavinathanson same here - neither of these is urgent, but wanted to let you know that the comments were addressed

cohorts/survival.py
```diff
@@ -145,41 +122,185 @@ def plot_kmf(df,
...
+          defaults to "condition_name = condition_value", or *_condition_label
+          options in case of boolean operators.
+        color_palette: str, optional, name of sb.color_palette to use
+          if color_map not provided.
+        print_as_title: bool, optional, whether or not to print text
+          within the plot's title vs. stdout, default False
+    """
+    
+    # set reasonable default threshold value depending on type of condition_col
+    if threshold is None:
```
tavinathanson, line 242: Doesn't this duplicate logic in `plot_survival`? Should we move the `median` default out of there?

cohorts/survival.py
```diff
@@ -145,41 +122,185 @@ def plot_kmf(df,
...
+            df=df,
+            title=title,
+            ax=ax))
+        return _plot_kmf_single(**arglist)
+    else:
+        # prepare for stratified analysis
+        if threshold == 'median':
+            # by default, "median" threshold should be intra-strata median
+            arglist['threshold'] = df[condition_col].dropna().median()
+        elif threshold == 'median per-strata':
```
tavinathanson: Weird to have an argument with a space in it? Maybe `stratamedian`?

cohorts/survival.py
```diff
@@ -145,41 +122,185 @@ def plot_kmf(df,
...
+            arglist['threshold'] = 'median'
```
tavinathanson: Nit: consistency with `'` vs. `"`, replace with `"` per other `"`s below?

cohorts/survival.py
```diff
@@ -145,41 +122,185 @@ def plot_kmf(df,
...
+                                  if 'median', condition_col thresholded
+                                  at its median
+                                  if 'median per-strata', & if stratified analysis
+                                  then condition_col thresholded by strata
+        title: Title for the plot, default None
+        ax: an existing matplotlib ax, optional, default None
+             note: not currently supported when `strata_col` is not None
+        with_condition_color: str, hex code color for the with-condition curve
+        no_condition_color: str, hex code color for the no-condition curve
+        with_condition_label: str, optional, label for TRUE condition case
```
tavinathanson: Nit: `True` in Python vs. `TRUE` in R

tavinathanson (approved):
LGTM! Only minor comments, feel free to merge whenever without another review

PR #241, opened 2017-07-24
Expose gcloud utils

jburos:
Currently, gcloud utils cannot be imported because they weren't included in setup.py. This PR includes them for use in certain projects.


tavinathanson (approved):
LGTM!

PR #242, opened 2017-07-24
add `join_on_left` parameter to dataframe_loader

jburos:
Allows a dataframe to merge with the cohort on a field other than patient_id.

In one of our projects, the extraneous data are sometimes labelled by sample_id, which is a different identifier than patient id. Allowing a join_on_left parameter seems cleaner than merging-in a patient_id identifier for each of these external dataframes.

(I'm not really happy with the name of this parameter, but the functionality would be useful).

### DIFF
--- a/cohorts/dataframe_loader.py
+++ b/cohorts/dataframe_loader.py
@@ -28,7 +28,7 @@ class DataFrameLoader(object):
         The column of the `DataFrame` to join on (i.e. the patient
         ID column name).
     join_on_left: str
-        The corresponding column in the `cohorts.Patient.additional_data` 
+        The corresponding column in the `cohorts.Patient.additional_data`
 	to join on, if not the `id` (which is the default).
     """
     def __init__(self,
@@ -39,7 +39,7 @@ class DataFrameLoader(object):
                  join_on_left="patient_id"):
         self.name = name
         self.load_dataframe = load_dataframe
-        if join_on:
+        if join_on is not None:
             warnings.warn("`join_on` parameter is deprecated. Please use `join_on_right` instead.", DeprecationWarning)
             self.join_on_right = join_on
         else:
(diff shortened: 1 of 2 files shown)


cohorts/dataframe_loader.py
```diff
@@ -29,7 +29,9 @@ class DataFrameLoader(object):
     def __init__(self,
                  name,
                  load_dataframe,
-                 join_on="patient_id"):
+                 join_on="patient_id",
```
tavinathanson: Rename this to `join_on_right`? Otherwise I think it's confusing.

If you're worried about backwards compat., can add an additional `join_on` parameter defaulted to `None`; if passed in, log a https://docs.python.org/3/library/exceptions.html#DeprecationWarning and pass to `join_on_right`?

tavinathanson (changes requested):
LGTM! I feel like there was a reason I only did `patient_id`, but can't remember. But if this is working for you, great!

Just a minor comment.

jburos:
@tavinathanson made what i think are the requested changes on this - do you mind taking a look?

cohorts/dataframe_loader.py
```diff
@@ -22,14 +24,24 @@ class DataFrameLoader(object):
...
                  name,
                  load_dataframe,
-                 join_on="patient_id"):
+                 join_on=None,
+                 join_on_right="patient_id",
+                 join_on_left="patient_id"):
         self.name = name
         self.load_dataframe = load_dataframe
-        self.join_on = join_on
+        if join_on:
```
tavinathanson: Nit: I think `if join_on is not None` is less confusing, since `if join_on` would jump to the else if `join_on` is `""`, `False`, etc. https://stackoverflow.com/questions/7816363/if-a-vs-if-a-is-not-none

From PEP:

> Also, beware of writing if x when you really mean if x is not None -- e.g. when testing whether a variable or argument that defaults to None was set to some other value. The other value might have a type (such as a container) that could be false in a boolean context!
https://www.python.org/dev/peps/pep-0008/#id51

In this case I don't think it matters, but just a general comment :)

tavinathanson (approved):
LGTM!

Issue #247, opened 2017-08-22
KeyError when predicting neoantigens


jburos:
Encountered an error when using `expressed_neoantigen_count` in the course of an analysis:

This did not occur for all patients, but did occur for approximately 3 patients. 

Here is a sample of the relevant output:
```
expressed_neoantigen_count:  18%|████████████████████████████████████▋                                                                                                                                                                          | 14/79 [01:04<12:12, 11.27s/it]
Traceback (most recent call last):
  File "/demeter/users/burosj01/projects/netmhc-bundle/netMHC-3.4/netMHC.py", line 536, in <module>
    main()
  File "/demeter/users/burosj01/projects/netmhc-bundle/netMHC-3.4/netMHC.py", line 528, in main
    fulldict[allele] = printresults(NNpred, seqs, inorder, preds, allele, methodlist[allele], peplen, wwwrun)
KeyError: 'HLA-B07:02'
Traceback (most recent call last):
  File "/demeter/users/burosj01/projects/netmhc-bundle/netMHC-3.4/netMHC.py", line 536, in <module>
    main()
  File "/demeter/users/burosj01/projects/netmhc-bundle/netMHC-3.4/netMHC.py", line 528, in main
    fulldict[allele] = printresults(NNpred, seqs, inorder, preds, allele, methodlist[allele], peplen, wwwrun)
KeyError: 'HLA-B07:02'
Traceback (most recent call last):
  File "/demeter/users/burosj01/projects/netmhc-bundle/netMHC-3.4/netMHC.py", line 536, in <module>
    main()
  File "/demeter/users/burosj01/projects/netmhc-bundle/netMHC-3.4/netMHC.py", line 528, in main
    fulldict[allele] = printresults(NNpred, seqs, inorder, preds, allele, methodlist[allele], peplen, wwwrun)
KeyError: 'HLA-B07:02'
Traceback (most recent call last):
  File "/demeter/users/burosj01/projects/netmhc-bundle/netMHC-3.4/netMHC.py", line 536, in <module>
    main()
  File "/demeter/users/burosj01/projects/netmhc-bundle/netMHC-3.4/netMHC.py", line 528, in main
    fulldict[allele] = printresults(NNpred, seqs, inorder, preds, allele, methodlist[allele], peplen, wwwrun)
KeyError: 'HLA-B07:02'
Traceback (most recent call last):
[...]

Issue #246, opened 2017-08-22
logging does not respond to user-configured level


jburos:
The way our logging module is configured isn't quite right - I don't know what the best practice is (seems there are [multiple recommendations on SO](https://stackoverflow.com/questions/15727420/using-python-logging-in-multiple-modules)). 

At the very least, I find this problematic when trying to debug certain errors.  

For example, after adding various debug messages (e.g. [here](https://github.com/hammerlab/cohorts/commit/3f1a7e6836bb74e23e41650f5dcab5a425c0bc82)), I see the following:

```py
# (note that my cohort is defined here in the object `rcc`)
import logging
logging.getLogger('cohorts.cohort').setLevel(logging.DEBUG)
x = rcc.cohort._load_single_patient_neoantigens(rcc.cohort.patient_from_id('4_54'), 
    only_expressed=True, epitope_lengths=[8,9,10,11], 
    ic50_cutoff=500, process_limit=10, max_file_records=None,
    filter_fn=None)
```
The only output I see is a warning (also added in the same edit):

```
WARNING:cohorts.cohort:patient 4_54 has no expressed variants
```

I think this is in part because our [`get_logger` method](https://github.com/hammerlab/cohorts/blob/master/cohorts/utils.py#L203) sets the level to `logging.INFO` pro-actively. It also could be a result of the `logging.disable(logging.INFO)` line in [cohorts.cohort.get_filtered_isovar_epitopes](https://github.com/hammerlab/cohorts/blob/master/cohorts/cohort.py#L1099). This does not appear to have been reverted, and I'm also not sure what its purpose is.

At any rate, in general this leads to inconsistent behavior in the package. It would be helpful to figure out how to repair the logging & ensure that the user can capture relevant messages reliable.

Some resources I've been looking at include:

* [(docs) Configuring logging for a library](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library)
* [(SO) How should logging be used in a package?](https://stackoverflow.com/questions/27016870/how-should-logging-be-used-in-a-python-package)

PR #249, opened 2017-08-22
Small fixes to improve estimation of expressed neoantigens 

jburos:
Added a variety of debug messages, also modified isovar cache so that the cached file varies according to the input variants provided. In the absence of this, I was seeing the same number of expressed variants irrespective of which filter_fn was applied.

Finally, captured the case documented in #247 when trying to predict neoantigens using `cohorts.functions.expressed_neoantigen_count` when no variants were expressed.


cohorts/cohort.py
```diff
@@ -1021,29 +1025,36 @@ def _load_single_patient_neoantigens(self, patient, only_expressed, epitope_leng
...
-            # the variant later.
-            for variant_column in ["chr", "pos", "ref", "alt"]:
-                # Be consistent with Topiary's output of "start" rather than "pos".
-                # Isovar, on the other hand, outputs "pos".
-                # See https://github.com/hammerlab/topiary/blob/5c12bab3d47bd86d396b079294aff141265f8b41/topiary/converters.py#L50
-                df_column = "start" if variant_column == "pos" else variant_column
-                df_epitopes[df_column] = df_epitopes.source_sequence_key.apply(
-                    lambda key: dict(key)[variant_column])
-            df_epitopes["patient_id"] = patient.id
+            if len(df_isovar.index) == 0:
```
tavinathanson: Why not just `len(df_isovar)`?

cohorts/cohort.py
```diff
@@ -1088,9 +1099,11 @@ def get_filtered_isovar_epitopes(self, epitopes, ic50_cutoff):
         return EpitopeCollection(mutant_binding_predictions)
 
     def load_single_patient_isovar(self, patient, variants, epitope_lengths):
-        # TODO: different epitope lengths, and other parameters, should result in
-        # different caches
-        isovar_cached_file_name = "%s-isovar.csv" % self.merge_type
+        # different cache depending on epitope length & variant set
+        variant_hash = make_hash(frozenset(variants))
```
tavinathanson: Nice! Are we confident that this hashes to the same value for the same `VariantCollection`?

cohorts/hash.py
```diff
@@ -0,0 +1,42 @@
+import copy
+
+# courtesy of https://stackoverflow.com/a/8714242/3457743
+
+DictProxyType = type(object.__dict__)
+
+def make_hash(o):
```
tavinathanson: Style: lots of unnecessary newlines?

cohorts/hash.py
```diff
@@ -0,0 +1,42 @@
+import copy
+
+# courtesy of https://stackoverflow.com/a/8714242/3457743
```
tavinathanson: Can you throw this in the docstring and move the function to utils or something? Extra file seems unnecessary to me; but if you disagree I don't feel strongly.

tavinathanson (commented):
Looks great, thanks for fixing these!

One thing I'm not clear on is why this fixes the `filter_fn` issue. My understanding is that `_load_single_patient_neoantigens` filters by `filter_fn` at the very end, after loading from the cache; so what changed here that makes that work better?

cohorts/cohort.py
jburos: It does if 

1. you're on the same hardware/OS install
2. you `export PYTHONHASHSEED=0`

Re the PYTHONHASHSEED, I set this in my `~/.bashrc`.  

One thing i've thought about doing is alerting the user if this env variable hasn't been set. Curious to know what you think - is this overkill?

tavinathanson: Why not just use `hashlib.md5` to avoid both of these issues?

cohorts/hash.py
jburos: If you like. I don't have a strong preference re: more or fewer files, except when it helps organization.

tavinathanson: Cool, that's my preference then

jburos:
Thanks @tavinathanson! This is really helpful.

Re: the isovar change, I wasn't seeing this with the neoantigen case but I did see it when in an example I was working with - which calculated a weighted sum of variants where the weight was based on the level of expression. If you're interested in seeing the specific function, it's been pushed to a development branch [here](https://github.com/hammerlab/cohorts/blob/develop-rcc/cohorts/functions.py#L149-L205). 

I haven't gone back to look at how this might impact the filtering by expression -- I'm not sure it does, since there are "be careful with isovar the cache doesn't depend on parameters" messages throughout the code! -- but it does concern me that the cache may have been initialized with a partially-filtered set of variants, leading to underestimating expressed variants on subsequent calls with a more relaxed filter. If this cache is error-prone, we would probably want to restructure this function to only ever cache using an unfiltered set of variants & filter by variants later.

Definitely open to feedback since I want to make sure we get this right.

tavinathanson:
@jburos what do you mean by "that the cache may have been initialized with a partially-filtered set of variants"? Where does that happen?

cohorts/cohort.py
jburos: `md5` only works for strings -- not Variants. Unless you `pickle.dump` them, or do `str(x)` for each -- which is also kind of hacky.

tavinathanson: I agree that it's hacky, but sounds better to me than dealing with the above issues.

jburos:
@tavinathanson I don't think it would happen in our normal workflow, but we don't know what variants are passed into this function the first time it's called - it could be called directly by the user, with only `indels`. In that case the indels-only would be cached & used on subsequent calls (e.g. for `expressed_snv_count`).

Otherwise the only two places this is used within the code include load_neoantigens & the `variant_expressed_filter`, neither of which passes a filtered set of variants to `load_single_patient_isovar`.

tavinathanson:
@jburos gotcha, makes sense! Maybe we should put https://github.com/hammerlab/cohorts/blob/master/cohorts/cohort.py#L989 inside this function?

jburos:
Yep, that's what I was thinking as well. Then there's always an option to filter by the variants the user provides after reading from the cache (analogous to the way other cached-items work). 

But, would this mean you don't want to cache at all based on the variants or epitope-lengths given? Not really convinced we want to give up on that either (though, epitope-lengths could be done easily since they're numeric).

tavinathanson:
@jburos I think we can have both of those things. Caching on the `isovar` results is useful in terms of time savings, for sure; and this does it in a better way. And loading in unfiltered variants addresses other potential bugs.

jburos:
So, I tried to add a `filter_by_variants` step to the `load_single_patient_isovar` process (based on a portion of code from `isovar.allele_reads` [here](https://github.com/hammerlab/isovar/blob/master/isovar/allele_reads.py#L212-L227)), but on further thought it seems much cleaner to leave this logic within `isovar`. Instead I now have a phantom `_load_single_patient_isovar_unfiltered` function which isn't used anywhere. I will likely remove this. 

Also ended up tweaking our logging code (per #246) so that I could confirm that the cache-files were indeed being re-used without problems. LMK if you have a strong opinion about this. I can revert it, just needed it temporarily so that I could actually see my debug messages.

jburos:
@tavinathanson no rush, just FYI I'm done making updates on this PR for now.

cohorts/cohort.py
```diff
@@ -1087,20 +1097,39 @@ def get_filtered_isovar_epitopes(self, epitopes, ic50_cutoff):
...
-        isovar_cached_file_name = "%s-isovar.csv" % self.merge_type
-        df_isovar = self.load_from_cache(self.cache_names["isovar"], patient.id, isovar_cached_file_name)
+    def _load_single_patient_isovar_unfiltered(self, patient, epitope_lengths):
+        """
+        Version of _load_single_patient_isovar, runs on all variants without filtering
+        """
+        logger.debug("Loading unfiltered isovar data for patient: {}".format(patient.id))
+        epitope_hash = make_hash(frozenset(epitope_lengths))
+        cache_file_name = "{}-isovar.{}.csv".format(self.merge_type, str(epitope_hash))
+        import pdb; pdb.set_trace()
```
tavinathanson: Let's get rid of this PDB statement :)

tavinathanson (changes requested):
I think the two separate `isovar` functions are fairly confusing. Why not, as you mentioned, add a `filter_isovar` just like `filter_neoantigens`, basically copying the code in https://github.com/hammerlab/cohorts/blob/master/cohorts/varcode_utils.py#L134 and https://github.com/hammerlab/cohorts/blob/master/cohorts/varcode_utils.py#L47, e.g. a new `FilterableIsovar` class? And then only have only function like `_load_single_patient_neoantigens`?

jburos:
Well, I did that at first -- then undid it. It doesn't seem right to me to repeat the logic that's in isovar -- better to keep it in isovar (ie if the logic changes in isovar, wouldn't we then just repeat it here?).

As it stands, we never have a use case for querying on all variants -- in every case we pull the variants in before calling this function -- so if it's confusing I'd prefer to just remove the `_unfiltered` function & leave the current one as-is. What do you think -- does that make sense?

cohorts/cohort.py
jburos: oops!

Issue #253, opened 2017-08-26
all_effects caches, but does not return, all effects


tavinathanson:
Found when trying to fix https://github.com/hammerlab/cohorts/issues/252

When passing `all_effects` to `load_effects`, all the effects are cached, but then `effects` is further mutated before being returned. (All effects are returned when they're pulled from the cache on a later call, too.)

tavinathanson:
Fixed in #254

Issue #252, opened 2017-08-25
Don't cache after only top-priority effects; post-filter instead


tavinathanson:
Oof, I found what seems to be a gnarly bug in Cohorts. It shouldn't impact any conclusions we've made, but it does impacts effect/variant-with-effect counts.

Notebook with example: https://github.com/hammerlab/bladder-analyses/blob/master/analyses/notebooks/Surfacing%20a%20load_effects%20bug%20in%20Cohorts.ipynb

Up until [this](https://github.com/hammerlab/cohorts/pull/165) PR, we never saved all effects; only the top-priority effects. Currently, we do save all effects, but usually we only use the top-priority effects.

What if we have a Substitution that is lower priority than another effect, but we end up filtering to just Substitutions? I ran into this while trying to generate all missense SNVs in the bladder cohort ( (and verify # == `missense_snv_count`) for Alex S.

Note: I could see an argument for this not being a bug in some scenarios, when we only care about the top priority effect.

The solution seems clear: we need to apply top-priority filtering as a post-filter rather than caching it.

tavinathanson:
The good news is that `missense_*` and `nonsynonymous_*` functions use `only_nonsynonymous=True`, which is cached separately and helps a lot (it filters by priority *after* filtering by silent/noncoding).

This bug could explain why we sometimes find discrepancies between functions that should be equivalent.

tavinathanson:
Oops, this is worse than I thought. https://github.com/hammerlab/cohorts/pull/165 made it so `only_nonsynonymous` uses the effect collection that was previously filtered to top priority, and *then* filters further. That is definitely wrong.

tavinathanson:
Fixed in #254

Issue #155, opened 2016-10-14
Write doc "Customizing filter functions"


jburos:
Would be useful to have a doc explaining to users how to write their own filter functions.

PR #254, opened 2017-08-26
Fix effect priority caching

tavinathanson:
In this PR, fixes for: https://github.com/hammerlab/cohorts/issues/253 and https://github.com/hammerlab/cohorts/issues/252

Instead of an "all_effects" cache, we now remove priority filtering from the small nonsyn. cache and the larger effects cache. If `all_effects` is true, we don't do priority filtering after generating our collections or pulling from the cache; otherwise, we do.

Tried to test this pretty aggressively.

### DIFF
--- a/cohorts/cohort.py
+++ b/cohorts/cohort.py
@@ -826,14 +826,14 @@ class Cohort(Collection):
         if variants is None:
             return None
 
-        def top_priority_maybe(effects):
+        def top_priority_maybe(effect_collection):
             """
             Always (unless all_effects=True) take the top priority effect per variant
             so we end up with a single effect per variant.
             """
             if all_effects:
-                return effects
-            return EffectCollection(list(effects.top_priority_effect_per_variant().values()))
+                return effect_collection
+            return EffectCollection(list(effect_collection.top_priority_effect_per_variant().values()))
 
         if only_nonsynonymous:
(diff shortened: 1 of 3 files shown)


tavinathanson:
Note that we need to clear our effects caches to make use of this PR. Should also result in a non-trivial version bump.

jburos (approved):
This looks good -- thanks @tavinathanson. One small suggestion - why not move the `top_priority_maybe` filter into [varcode_utils.py](https://github.com/hammerlab/cohorts/blob/master/cohorts/varcode_utils.py), along with `filter_effects` & have it take the parameter for `all_effects`? And/or, move this logic within the filter-effects function? 

Down the line, I can see where it might be helpful to allow the user to customize when the de-duping will happen, which could also aid transparency .. but for now this looks great.

When we get to writing a how-to on functions #155  (will likely wait until the current refactoring is stable) we should mention that the de-duping always happens _last_. This would be relevant, for example, if one wanted to write a "not-X" filter function.

tavinathanson:
@jburos great idea! Just moved `top_priority_maybe`. Good to merge on green?

jburos:
@tavinathanson sorry if you didn't see my 👍  -- but yep good to go whenever you are ready. Once this has been merged I will update my pending PRs to be consistent with these changes.

PR #256, opened 2017-08-31
Fix effects filtering for splice sites

### LINKED ISSUE #255: ExonicSpliceSite with Subsitution/etc. alternate_effect is not included in filtering
From @tavinathanson:

> :thinking_face: is https://github.com/hammerlab/cohorts/blob/master/cohorts/functions.py#L156 (and others) incomplete because it's not counting `ExonicSpliceSite`  mutations that have `Subsitution` as the alternate effect?

From @iskandr:

> OK, looked at the source and the only properties that forward through the alternate effect are    mutant_protein_sequence, modifies_protein_sequence, modifies_coding_sequence
>So, you have to currently explicitly check, which seems like a recipe for making this mistake often
>Possible solutions: add a `is_substitution` method or change how  exonic splice sites are handled
>One possible alternative would be just having every Intragenic effect take a `splice_site` boolean attribute
>Or something else that better preserves the meaning of the class hierarchy

From @tavinathanson:
>yikes glad we caught it now at least
>how about we invert the meaning of `alternate_effect` and have every effect have an optional `alternate` where `ExonicSpliceSite` would be one such alternate?
>similar to `splice_site` boolean, which also seems fine to me
>i think either of those is better than `is_subsitution` which feels like a hack
>for now i'm going to hack around this in cohorts until we know exactly what we want in varcode

tavinathanson:
Until `varcode` handles this better, I've just added some logic to `filter_effects` that handles splice sites with alternate effects that we don't want to filter out + relevant tests.

Fixes #255


iskandr (approved):
LGTM but we should fix this more elegantly in Varcode

tavinathanson:
Evaluated impact on lung data here (sorry for private repo link): https://github.com/hammerlab/msk-bms-lung/blob/master/analyses/notebooks/_exploratory_notebooks/Compare%20Cohorts%20Versions%20on%20mini_data.py%20vs.%20New%20Splice%20Filtering.ipynb

